PR #3, opened 2016-02-27
Broken autocomplete, do not review right now

ehamp:
- added gmaps.js and Google maps as an asset 
- put a basic map onto the page


app/assets.py
```diff
@@ -24,3 +24,15 @@
...
+    'gmaps.js',
+    filters='jsmin',
+    output='scripts/gmaps.js'
+)
+
+googlemaps_js = Bundle(
+    'googlemaps.js',
+    filters='jsmin',
+    output='scripts/googlemaps.js'
+)
```
weronica: Add a new line here :)

weronica:
Since this is a front-end task, can you post a screenshot of your work?

PR #4, opened 2016-02-27
Daniel tables

weronica:


app/models/__init__.py
```diff
@@ -4,3 +4,4 @@
 """
 
 from user import *  # noqa
+from descriptor import *
```
armant: nit: let's arrange imports in alphabetical order.

app/models/descriptor.py
```diff
@@ -0,0 +1,44 @@
...
+    resource = db.relationship('Resource', back_populates='text_descriptors')
+    descriptor = db.relationship('Descriptor', back_populates='text_resources')
+
+class Descriptor(db.Model):
+    __tablename__ = 'descriptors'
+    id = db.Column(db.Integer, primary_key=True)
+    name = db.Column(db.String(64), index=True)
+    #category = db.Column(db.String(64), index=True)
+    values = db.Column(db.PickleType)
+    text_resources = db.relationship('TextAssociation', back_populates='descriptor')
```
armant: What is the use case for having this relationship? (Descriptor -> TextAssociation)
ditto for (descriptor -> OptionAssociation), as well as (Resource -> TextAssociation) and (Resource -> OptionAssociation). I understand other way around but not this way.

armant:
r @weronica. @dzhang55 please see comments.

app/models/descriptor.py
dzhang55: I thought a bidirectional relationship would be useful because if you were to filter resources by a given descriptor, you could find the descriptor and then use the association to get all resources with that descriptor. On the other hand, if you click on a pin you would go the other way as you would have a resource and then use the association to get all the descriptors for that resource.

PR #11, opened 2016-03-15
Add skeletons for descriptor CRUD, resource CRUD, and suggestions

armant:


app/resource/views.py
```diff
@@ -0,0 +1,8 @@
+from flask import render_template
+from flask.ext.login import login_required
+from . import resource
+
+@resource.route('/')
+@login_required
+def view():
```
weronica: Let's call the function for the default route `index`. The corresponding template should be called `resource/index.html`.

app/suggestion/views.py
```diff
@@ -0,0 +1,8 @@
+from flask import render_template
+from flask.ext.login import login_required
+from . import suggestion
+
+@suggestion.route('/')
+@login_required
+def view():
```
weronica: Same as above: `index`/`index.html`.

app/descriptor/views.py
```diff
@@ -0,0 +1,8 @@
+from flask import render_template
+from flask.ext.login import login_required
+from . import descriptor
+
+@descriptor.route('/')
+@login_required
+def view():
```
weronica: Let's call the function for the default route `index`. The corresponding template should be called `descriptor/index.html`.

app/__init__.py
```diff
@@ -68,4 +68,13 @@ def create_app(config_name):
     from bulk import bulk as bulk_blueprint
     app.register_blueprint(bulk_blueprint, url_prefix='/bulk')
 
+    from descriptor import descriptor as descriptor_blueprint
+    app.register_blueprint(descriptor_blueprint, url_prefix='/descriptor')
+
+    from resource import resource as resource_blueprint
```
weronica: If I'm understanding correctly, the `resource` blueprint is for making changes to single resources, while the `bulk` blueprint is for making changes to many resources at once. Seems like they should have similar names? Perhaps they should be called `individual_resource` and `bulk_resource`? Or maybe they actually aren't that similar? What do you think?

app/__init__.py
```diff
@@ -68,4 +68,13 @@ def create_app(config_name):
     from bulk import bulk as bulk_blueprint
     app.register_blueprint(bulk_blueprint, url_prefix='/bulk')
 
+    from descriptor import descriptor as descriptor_blueprint
```
weronica: Ok I've thought more about this now. Here is a dump of my thoughts...

My concern with the phrase "resources have descriptors" is that it seems like each resource could have its own set of descriptors; however, in reality, there is a global set of descriptors that every resource "responds" to. So the correct phrase is something more like "resources have descriptor_responses" or something. How does that sound? 

To be clear, the `descriptor` blueprint will contain functionality for changing the descriptors themselves, instead of changing individual resources' descriptor_responses? 

(I'm not proposing an immediate code change; I'm interested in starting a dialogue about what we're naming things.)

app/__init__.py
```diff
@@ -68,4 +68,13 @@ def create_app(config_name):
...
+    app.register_blueprint(descriptor_blueprint, url_prefix='/descriptor')
+
+    from resource import resource as resource_blueprint
+    app.register_blueprint(resource_blueprint, url_prefix='/resource')
+
+    from suggestion import suggestion as suggestion_blueprint
```
weronica: Will the `suggestion` blueprint contain functionality both to publicly suggest new resources and to (as an admin) approve/deny suggested resources? Or just the latter? Personally, I think it makes sense to include both in this blueprint.

weronica:
I left a few comments, which are probably a jumping point for an in-person conversation. We can chat tonight.

app/descriptor/views.py
armant: Yeah, I've thought about it too, but it seems like a bad style to have more than one index. What do you think about that?

app/__init__.py
armant: For now I think it's the latter, because it seems to me that all external users' actions (viewing the map & suggestions) should be in one blueprint, `external_user` or something like that. But I guess we don't have to make that decisions right now anyway.

app/__init__.py
armant: Yeah, I see what you mean - it's more natural to think that one descriptor describes one specific thing. But between `descriptor` and `descriptor_response` I would still choose the former, because it's also unnatural to think about something responding to a description. I'd say let's leave it as it is unless we find a more intuitive alternative. The inconsistency in the meaning doesn't seem to be very big to me in this case, especially if we document everything clearly.

Do you think it's a big concern that an eponymous model exists? It doesn't seem so to me, since both are meant to mean the same thing.

app/__init__.py
weronica: Hmm I'm not sure that's a good enough reason to not have the former functionality in this blueprint. `@login_required` is specified at the route level and not at the blueprint level, so we can have a mixture of routes with and without `@login_required` in a single blueprint.

app/__init__.py
armant: Yeah, individual_resource and bulk_resource sound better, let's do that.

app/descriptor/views.py
weronica: There isn't actually more than one `index` - there's `main.index`, `admin.index`, `descriptor.index`... They're all different :) The naming convention `index` just indicates that this is the "main" page for that blueprint.

armant: Sure, that makes sense.

app/__init__.py
weronica: I'm not proposing that the blueprint be called `descriptor_response` - I'm just clarifying how we should talk about the relationship between resources and descriptors. Sorry for the confusion. 

I do think it's confusing that an eponymous model exists, since a blueprint should be at a higher level of abstraction than any one model. What about the name `schema`?

armant: Hmm, sounds too vague to me.

app/__init__.py
weronica: or perhaps `individual_crud` and `bulk_crud` (lol at the name, but I think it's descriptive)

armant: I'd say let's keep it with 'resource', just in case we will add more functionality (other than CRUD) later?

weronica: What else is there besides CRUD?

armant: I mean not right now, but possibly later.
Also bulk_crud doesn't make much sense, since U won't be done in bulk, and R is always in bulk.

PR #14, opened 2016-03-25
CSV model, row name suggestions

weronica:


armant:
r @dzhang55

app/bulk_resource/views.py
```diff
@@ -1,8 +1,11 @@
 import csv
```
armant: nit: newline after this

app/bulk_resource/views.py
```diff
@@ -1,8 +1,11 @@
...
+from datetime import datetime
 from flask import redirect, render_template, url_for
```
armant: nit: newline after this

app/bulk_resource/views.py
```diff
@@ -1,8 +1,11 @@
...
 from flask.ext.login import login_required
 from . import bulk_resource
```
armant: nit: move ".", "..", and others start start with a symbol to before the first one that starts with a letter

app/bulk_resource/views.py
weronica: Done

app/bulk_resource/views.py
weronica: Done

app/admin/views.py
```diff
@@ -1,18 +1,17 @@
...
+from . import admin
+from .. import db
+from ..decorators import admin_required
+from ..email import send_email
+from ..models import Role, User
 from forms import (
     ChangeUserEmailForm,
     NewUserForm,
     ChangeAccountTypeForm,
     InviteUserForm,
```
dzhang55: Can this be alphabetical order? (also I think I actually fix this in my PR so hopefully that's okay)

app/bulk_resource/views.py
```diff
@@ -25,4 +48,6 @@ def upload():
 @bulk_resource.route('/review')
 @login_required
 def review():
-    return render_template('bulk_resource/review.html')
+    csv_container = CsvContainer.query.first()
```
dzhang55: This would also have to be associated with the current_user right?

app/admin/views.py
weronica: Done

app/bulk_resource/views.py
weronica: Yes, definitely. For now, I added a `TODO`. Something is broken with `current_user` at the moment...

PR #15, opened 2016-03-25
Loaded pins onto the map with a mouse over info window

ehamp:
loaded pins from database onto map
added a hover over info window that still needs to have the name added

### DIFF
--- a/app/assets/scripts/vendor/map.js
+++ b/app/assets/scripts/vendor/map.js
@@ -93,14 +93,13 @@
         }).done(function(data){
             data = JSON.parse(data)
             for(var i = 0; i < data.length; i++){
-            //we have the pins data, access it here and place pins on map
-              latLng = new google.maps.LatLng(data[i].Latitude, data[i]
-              .Longitude)
+              latLng = new google.maps.LatLng(data[i].latitude, data[i]
+              .longitude)
               var marker = new google.maps.Marker({
                 position: latLng,
                 map: map,
                 visible: true,
-                title: data[i].Name
+                title: data[i].name
               });
 
               var infowindow = new google.maps.InfoWindow({
--- a/app/main/views.py
+++ b/app/main/views.py
@@ -9,13 +9,14 @@ def index():
     return render_template('main/index.html')
 
 @main.route('/get-resource', methods=['GET'])
-def getResource():
+def get_resource():
     Resource.generate_fake()
     pins = Resource.query.all()
     data = []
     for pin in pins:
-       this_pin = {'Name': pin.name, 'Latitude': pin.latitude, 'Longitude':
-           pin.longitude}
+       this_pin = {
+           'name': pin.name,
+           'latitude': pin.latitude,
+           'longitude': pin.longitude}
        data.append(this_pin)
-    print data
     return json.dumps(data)
--- a/app/main/views.py
+++ b/app/main/views.py
@@ -15,9 +15,9 @@ def get_resource():
     pins = Resource.query.all()
     data = []
     for pin in pins:
-       this_pin = {
+        this_pin = {
            'name': pin.name,
            'latitude': pin.latitude,
            'longitude': pin.longitude}
-       data.append(this_pin)
+        data.append(this_pin)
     return json.dumps(data)
--- a/app/main/views.py
+++ b/app/main/views.py
@@ -2,11 +2,12 @@ from flask import render_template
 from . import main
 from ..models import Resource
 import json
-from flask import jsonify
+
 
 @main.route('/', methods=['GET'])
 def index():
     return render_template('main/index.html')
+
 
 @main.route('/get-resource', methods=['GET'])
 def get_resource():
--- a/app/models/resource.py
+++ b/app/models/resource.py
@@ -1,5 +1,5 @@
-from flask import current_app
 from .. import db
+
 
 class OptionAssociation(db.Model):
     """Association between a resource and a descriptor with an index
@@ -84,7 +84,7 @@ class Resource(db.Model):
 
         for i in range(num_options):
             options.append(Descriptor(
-                name=fake.word(), 
+                name=fake.word(),
                 values=['True', 'False']
             ))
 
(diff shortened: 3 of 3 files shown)


armant:
r @nnarang7

app/assets/scripts/vendor/map.js
```diff
@@ -74,4 +74,35 @@
...
         setupClickListener('changetype-address', ['address']);
         setupClickListener('changetype-establishment', ['establishment']);
+
+        $.ajax({
+          type: "GET",
+          url: "/get-resource"
+        }).done(function(data){
+            data = JSON.parse(data)
+            for(var i = 0; i < data.length; i++){
+            //we have the pins data, access it here and place pins on map
```
weronica: Nit: capitalize comment and end with a period. `// We have the pins data. Access it here and place pins on map.`

app/main/views.py
```diff
@@ -1,7 +1,21 @@
...
+from flask import jsonify
 
-
-@main.route('/')
+@main.route('/', methods=['GET'])
 def index():
     return render_template('main/index.html')
+
+@main.route('/get-resource', methods=['GET'])
+def getResource():
```
weronica: nit: naming convention - `get_resource` instead of `getResource` (the former is more pythonic)

app/main/views.py
```diff
@@ -1,7 +1,21 @@
...
+    Resource.generate_fake()
```
weronica: This probably shouldn't be here, right? I think we'd first call `add_fake_data` (which would call `Resource.generate_fake()`).

app/main/views.py
```diff
@@ -1,7 +1,21 @@
...
+    pins = Resource.query.all()
+    data = []
+    for pin in pins:
+       this_pin = {'Name': pin.name, 'Latitude': pin.latitude, 'Longitude':
+           pin.longitude}
+       data.append(this_pin)
+    print data
```
weronica: Nit: remove this print statement :)

app/main/views.py
```diff
@@ -1,7 +1,21 @@
...
 def index():
     return render_template('main/index.html')
+
+@main.route('/get-resource', methods=['GET'])
+def getResource():
+    Resource.generate_fake()
+    pins = Resource.query.all()
+    data = []
+    for pin in pins:
+       this_pin = {'Name': pin.name, 'Latitude': pin.latitude, 'Longitude':
```
weronica: Can we format as follows?
```
pin_dict = {
   `Name`: pin.name, 
   `Latitude`: pin.longitude,
   `Longitude`: pin.longitude
}
```

app/models/resource.py
```diff
@@ -131,7 +131,13 @@ def print_resources():
             print resource.text_descriptors
             print resource.option_descriptors
 
-
+    @staticmethod
+    def get_resources():
+        resources = []
+        for resource in db.session.query(Resource).all():
+            print resource
```
weronica: Nit: remove this print statement :)

app/models/resource.py
```diff
@@ -131,7 +131,13 @@ def print_resources():
             print resource.text_descriptors
             print resource.option_descriptors
 
-
+    @staticmethod
+    def get_resources():
+        resources = []
+        for resource in db.session.query(Resource).all():
```
weronica: Instead of `db.session.query(Resource).all()`, can we use `Resource.query.all()` (to conform with other db calls)?

app/models/resource.py
```diff
@@ -131,7 +131,13 @@ def print_resources():
             print resource.text_descriptors
             print resource.option_descriptors
 
-
+    @staticmethod
+    def get_resources():
```
weronica: Actually I'm confused why we need this method. `Resource.query.all()` returns a list of all the `Resource` objects.

app/templates/main/database_print.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
```
weronica: What is this file for?

app/assets/scripts/vendor/map.js
```diff
@@ -87,6 +87,36 @@
...
         setupClickListener('changetype-establishment', ['establishment']);
 
+        $.ajax({
+          type: "GET",
+          url: "/get-resource"
+        }).done(function(data){
+            data = JSON.parse(data)
+            for(var i = 0; i < data.length; i++){
+            //we have the pins data, access it here and place pins on map
+              latLng = new google.maps.LatLng(data[i].Latitude, data[i]
```
nnarang7: Style nit: Latitude and Longitude should be lowercase names

app/assets/scripts/vendor/map.js
```diff
@@ -87,6 +87,36 @@
...
+              .Longitude)
+              var marker = new google.maps.Marker({
+                position: latLng,
+                map: map,
+                visible: true,
+                title: data[i].Name
```
nnarang7: Style nit: Name should be lowercase as well

app/assets/scripts/vendor/map.js
```diff
@@ -87,6 +87,36 @@
...
+              });
+
+              var infowindow = new google.maps.InfoWindow({
```
nnarang7: If it's easier for you: each marker doesn't have to have a separate infowindow show up when hovered over...you can just have mouseover "alt text" descriptions instead

app/models/resource.py
```diff
@@ -131,7 +131,13 @@ def print_resources():
             print resource.text_descriptors
```
nnarang7: Nit: remove print statements

app/models/resource.py
nnarang7: @weronica that's my bad..i made this method before: i'll remove it in my next PR

app/templates/main/database_print.html
nnarang7: I think it's a test html for printing the json dump

nnarang7:
looks good to me!

PR #23, opened 2016-04-08
Created suggestions model

nnarang7:
- Added suggestions model with id, resource_id, suggestions text, and suggestion type
- Added faker data method
- Added test cases for all three types of suggestions


armant:
r @weronica

app/models/suggestion.py
```diff
@@ -0,0 +1,79 @@
+from .. import db
+
+class Suggestion(db.Model):
+	"""
+	Association between a resource and potential suggestions for it
+	"""
+	__tablename__ = 'suggestions'
+	id = db.Column(db.Integer, primary_key = True)
+	#suggestion_type is 0 for insertion, 1 for edit, and 2 for deletion
+	suggestion_type = db.Column(db.Integer)
```
armant: do we need this field? Won't admin's actions be the same regardless of the type (specifically, take a look at the suggestion and then making necessary changes)?

armant:
Also, just realized we might need a timestamp for each suggestion :/ Let's discuss tomorrow.

armant:
Also read/unread

app/models/__init__.py
```diff
@@ -6,3 +6,4 @@
 from bulk import *  # flake8: noqa
 from resource import * # flake8: noqa
 from user import *  # flake8: noqa
+from suggestion import *
```
weronica: nit: Add `# flake8: noqa`

app/models/suggestion.py
```diff
@@ -0,0 +1,79 @@
+from .. import db
+
+class Suggestion(db.Model):
```
weronica: We perhaps also need some info about the person who is adding the suggestion (name, email address, phone number, etc.)

nnarang7:
screen shot 2016-04-17 at 2 48 15 pm

app/models/suggestion.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+class Suggestion(db.Model):
+    """
+    Association between a resource and potential suggestions for it
+    """
+    __tablename__ = 'suggestions'
+    id = db.Column(db.Integer, primary_key=True)
+    # suggestion_type is 0 for insertion, 1 for edit, and 2 for deletion
+    suggestion_type = db.Column(db.Integer)
```
armant: remove suggestion_type; to distinguish suggestions for new resources, we'll make resource_id of these suggestion -1 (and please add comment for that)

app/models/suggestion.py
```diff
@@ -0,0 +1,120 @@
...
+    resource_id = db.Column(db.Integer)
```
weronica: Can this be an actual `relationship` to a resource?

app/models/suggestion.py
```diff
@@ -0,0 +1,120 @@
...
+    suggestion_text = db.Column(db.String(150))
+    # 0 stands for read, 1 stands for unread
+    read = db.Column(db.Integer, default=0)
```
weronica: Can this be a boolean?

app/models/suggestion.py
```diff
@@ -0,0 +1,120 @@
...
+class Suggestion(db.Model):
+    """
+    Association between a resource and potential suggestions for it
+    """
+    __tablename__ = 'suggestions'
+    id = db.Column(db.Integer, primary_key=True)
+    # suggestion_type is 0 for insertion, 1 for edit, and 2 for deletion
+    suggestion_type = db.Column(db.Integer)
+    resource_id = db.Column(db.Integer)
+    suggestion_text = db.Column(db.String(150))
```
weronica: Perhaps a bigger upper bound? 150 characters is not very many.

app/models/suggestion.py
```diff
@@ -0,0 +1,120 @@
...
+    # 0 stands for read, 1 stands for unread
+    read = db.Column(db.Integer, default=0)
+    timestamp = db.Column(db.DateTime)
+    contact_name = db.Column(db.String(64))
+    contact_email = db.Column(db.String(64), unique=True, index=True)
```
weronica: Why `unique=True` and `index=True`? What if the same person makes multiple suggestions and uses the same email address?

app/suggestion/views.py
```diff
@@ -1,12 +1,15 @@
...
 from . import suggestion
+from ..models import Suggestion
 
 
 @suggestion.route('/')
 @login_required
 def index():
-    return render_template('suggestion/index.html')
+    suggestions = Suggestion.query.all()
+    print(suggestions)
```
weronica: Remove the `print` statement?

app/templates/suggestion/index.html
```diff
@@ -1,7 +1,42 @@
 {% extends 'layouts/base.html' %}
-
+<!--should we have a "mark as read" column in the table -->
```
weronica: Turn this into a `TODO` statement? This is a good idea for the future :)

app/templates/suggestion/index.html
```diff
@@ -1,7 +1,42 @@
...
+          <th>Timestamp</th>
+          <th>Suggestion</th>
+          <th>Contact Info</th>
+          <th>Action</th>
+        </tr>
+      </thead>
+      <tbody>
+        {% for suggestion in suggestions %}
+          <tr>
+            <td>{{ suggestion['timestamp'] }}</td>
```
weronica: Suggested format: `suggestion.timestamp` instead of `suggestion['timestamp']`. Even though both are valid, the former doesn't require using strings, which are error-prone.

app/templates/suggestion/index.html
```diff
@@ -1,7 +1,42 @@
...
+            <td>{{ suggestion['suggestion_text'] }}</td>
+            <td>{{ suggestion['contact_name'] }}
+              <br>
+              {% if suggestion['contact_email'] == null or suggestion['contact_email'] == ''%}
```
weronica: More pythonic way: `if suggestion.contact_email is None`

app/templates/suggestion/index.html
```diff
@@ -1,7 +1,42 @@
...
+                {{ suggestion['contact_email'] }}
+                <br>
+                {{ suggestion['contact_number'] }}
+              {% endif %}
+            </td>
+            <td>
+              {% if suggestion['resource_id'] == -1 %}
+                <a href="{{ url_for('single_resource.create') }}">Insert</a>
+              {% else %}
+                <a href="{{ url_for('single_resource.edit', resource_id=suggestion['resource_id']) }}">Edit</a>
```
weronica: Same as above: `suggestion.resource_id` instead

requirements.txt
```diff
@@ -1 +1,2 @@
+pytz
```
weronica: Can you add this to `requirements/common.txt` instead? This `requirements.txt` file is just used for deploying websites - we shouldn't need to add anything here.

app/templates/suggestion/index.html
```diff
@@ -1,7 +1,42 @@
...
+          <th>Suggestion</th>
+          <th>Contact Info</th>
+          <th>Action</th>
+        </tr>
+      </thead>
+      <tbody>
+        {% for suggestion in suggestions %}
+          <tr>
+            <td>{{ suggestion['timestamp'] }}</td>
+            <td>{{ suggestion['suggestion_text'] }}</td>
```
weronica: Same as above

app/templates/suggestion/index.html
```diff
@@ -1,7 +1,42 @@
...
+            <td>{{ suggestion['contact_name'] }}
```
weronica: Same as above.

app/models/suggestion.py
```diff
@@ -0,0 +1,120 @@
...
+            s_edit = Suggestion(suggestion_type=1, resource_id=r_added.id, suggestion_text=s_text,
+                                read=s_read, timestamp=s_timestamp, contact_name=s_contact_name,
+                                contact_email=s_contact_email, contact_number=s_contact_number)
+            db.session.add(s_edit)
+            try:
+                db.session.commit()
+            except IntegrityError:
+                db.session.rollback()
+
+    @staticmethod
```
armant: I guess this is not needed if we no longer have suggestion types?

tests/test_suggestions_model.py
```diff
@@ -0,0 +1,92 @@
...
+        self.app_context.pop()
+
+    def test_insertion_suggestion(self):
+        """Test suggestion of an insertion"""
+        s_text = "Name: Taco Time \n Address: 123 45th Street"
+        s_contact_name = "John Doe"
+        s_contact_number = "123-456-7890"
+        s_contact_email = "taco@time.com"
+        s_timestamp = datetime.now(pytz.timezone('US/Eastern'))
+        suggestion = Suggestion(suggestion_type=0, resource_id=-1, suggestion_text=s_text,
```
armant: will have to edit since no suggestion_type

tests/test_suggestions_model.py
```diff
@@ -0,0 +1,92 @@
...
+                                read=0, contact_name=s_contact_name, contact_email=s_contact_email,
+                                contact_number=s_contact_number, timestamp=s_timestamp)
+        db.session.add(suggestion)
+        db.session.commit()
+
+        r_in_table = Suggestion.query.filter_by(suggestion_type=0).first()
+        self.assertTrue(r_in_table is not None)
+        self.assertTrue(r_in_table.suggestion_text == s_text)
+        self.assertTrue(r_in_table.suggestion_type == 0)
```
armant: same

app/models/suggestion.py
nnarang7: good point: i'll get rid of that

app/models/suggestion.py
```diff
@@ -0,0 +1,119 @@
...
+import pytz
+
+
+class Suggestion(db.Model):
+    """
+    Association between a resource and potential suggestions for it
+    """
+    __tablename__ = 'suggestions'
+    id = db.Column(db.Integer, primary_key=True)
+    # resource_id is -1 for insertion
```
weronica: We don't need this comment anymore. If there is no resource, then the relationship will be `null`.

app/models/suggestion.py
```diff
@@ -0,0 +1,119 @@
...
+    resource_id = db.Column(db.Integer, db.ForeignKey('resources.id'))
+    suggestion_text = db.Column(db.String(250))
+    # 0 stands for read, 1 stands for unread
+    read = db.Column(db.Boolean, default=False)
```
weronica: How does this field if multiple users have access to the same suggestion? Will it be marked `read` for all users if one user looks at it?

app/models/suggestion.py
```diff
@@ -0,0 +1,119 @@
...
+    timestamp = db.Column(db.DateTime)
```
weronica: Could we call this `submitted_timestamp` (or something else capturing the notion of "submission") for additional clarity?

app/models/suggestion.py
```diff
@@ -0,0 +1,119 @@
...
+    contact_name = db.Column(db.String(64))
+    contact_email = db.Column(db.String(64))
+    contact_number = db.Column(db.String(64))
```
weronica: Could this be `contact_phone_number`?

requirements/common.txt
```diff
@@ -30,3 +30,4 @@ webassets==0.10.1
 Werkzeug==0.10.4
 wheel==0.24.0
 WTForms==2.0.2
+pytz
```
weronica: Can you re-export this list using `pip freeze > requirements/common.txt`? That way, the list will be in alphabetical order :)

tests/test_suggestions_model.py
```diff
@@ -0,0 +1,104 @@
...
+        db.create_all()
+
+    def tearDown(self):
+        db.session.remove()
+        db.drop_all()
+        self.app_context.pop()
+
+    def test_insertion_suggestion(self):
+        """Test suggestion of an insertion"""
+        s_text = "Name: Taco Time \n Address: 123 45th Street"
```
weronica: lol 😂

app/models/suggestion.py
```diff
@@ -0,0 +1,119 @@
+from .. import db
+from datetime import datetime
+import pytz
```
weronica: These 2 imports should go before `from .. import db` since they are python packages.

app/models/resource.py
```diff
@@ -71,6 +71,7 @@ class Resource(db.Model):
     __tablename__ = 'resources'
     id = db.Column(db.Integer, primary_key=True)
     name = db.Column(db.String(64), index=True)
+    suggestions = db.relationship('Suggestion', backref='resource', uselist=True)
```
armant: (very) nit: can you move this under `option_descriptors`, please? I think it'll preserve the order of vitality of fields to a resource :)

app/models/suggestion.py
armant: This will be implemented by DZ this week but yeah. I think the idea is that whoever reads it, will take action on it, so it's fine.

app/models/suggestion.py
armant: maybe `submission_time`

app/models/suggestion.py
armant: or `contact_phone` :)

app/models/suggestion.py
```diff
@@ -0,0 +1,119 @@
...
+
+        num_words = 10
+        for i in range(count):
+            s_text = fake.sentence(nb_words=num_words)
+            s_read = 1
+            s_timestamp = datetime.now(pytz.timezone('US/Eastern'))
+            s_contact_name = fake.word()
+            s_contact_email = fake.word() + "@" + fake.word() + ".com"
+            s_contact_number = "123-456-7890"
+            s_insert = Suggestion(resource_id=-1, suggestion_text=s_text,
```
armant: I think `resource_id` should just be omitted - then it'll create a `Suggestion` without a relationship to a `Resource`.

app/models/suggestion.py
```diff
@@ -0,0 +1,119 @@
...
+                                read=s_read, timestamp=s_timestamp, contact_name=s_contact_name,
+                                contact_email=s_contact_email, contact_number=s_contact_number)
+            db.session.add(s_edit)
+            try:
+                db.session.commit()
+            except IntegrityError:
+                db.session.rollback()
+
+    @staticmethod
+    def generate_fake_deletions(count=20):
```
armant: It doesn't seem like we need this method; isn't it functionally the same with the one right above?

tests/test_suggestions_model.py
```diff
@@ -0,0 +1,104 @@
+import unittest
+from app import create_app, db
+from app.models import Resource, Suggestion
+from sqlalchemy.exc import IntegrityError
+from datetime import datetime
+import pytz
```
armant: nit: pls move this to above `unittest`, since both of these are Python packages and have higher precedence over external packages and this project's imports. Pls also add a newline between `unittest` and `from app import create_app, db` to signify the difference between Python packages and this project's imports.

tests/test_suggestions_model.py
```diff
@@ -0,0 +1,104 @@
...
+        self.assertTrue(r_in_table is not None)
+        self.assertTrue(r_in_table.suggestion_text == s_text)
+        self.assertTrue(r_in_table.resource_id == r_added.id)
+        self.assertTrue(r_in_table.read == 1)
+        self.assertTrue(r_in_table.contact_number == s_contact_number)
+        self.assertTrue(r_in_table.timestamp is not None)
+        self.assertTrue(r_in_table.contact_name == s_contact_name)
+        self.assertTrue(r_in_table.contact_email == s_contact_email)
+
+    def test_delete_suggestion(self):
```
armant: No need for this; pls see comments above

PR #28, opened 2016-04-14
Added names and descriptions to info windows on click

ehamp:
- added names to info window
- added descriptions to info window
- added hover over name display 
- Note: future goal to add some sort of logic that prevents the user from opening too many info windows at once, so the screen doesn't get cluttered


armant:
r @weronica

PR #26, opened 2016-04-09
Ajax Requests to get Address 

ehamp:
- Loaded pins onto map
- Display window with address of resource appears on click through ajax requests 

Note: All these weird bower files seemed to appear from installing async. Unsure why.

PR #30, opened 2016-04-18
Refactoring map.js, main/views.py, removing /node-modules, fixing req…

armant:
…uirements


armant:
r @weronica @ehamp

app/main/views.py
```diff
@@ -1,31 +1,27 @@
...
+    for dict in resources_as_dicts:
+        del dict['_sa_instance_state']
+    return json.dumps(resources_as_dicts)
 
 
-@main.route('/get-info', methods=['POST'])
-def getInfo():
-    pin_info = Resource.query.filter_by(name=request.form['data']).first()
-    return json.dumps({'Address': pin_info.address})
+@main.route('/get-associations/<int:resource_id>')
```
weronica: Why is this no longer a `POST` request?

app/main/views.py
```diff
@@ -1,31 +1,27 @@
...
-        counter = counter + 1
-        data.append(this_pin)
-    return json.dumps(data)
+@main.route('/get-resources')
+def get_resources():
+    resources = Resource.query.all()
+    resources_as_dicts = [resource.__dict__ for resource in resources]
+    # .__dict__ returns the SQLAlchemy object as a dict, but it also adds a
+    # field '_sa_instance_state' that we don't need, so we delete it.
+    for dict in resources_as_dicts:
```
weronica: `dict` is a data type, so for safety and clarify we probably shouldn't use it as a variable type. Perhaps just `d` instead?

weronica: Also, might be worth writing a helper method on the `Resource` model that returns the contents as a `dict` directly.

requirements/common.txt
```diff
@@ -13,20 +10,15 @@ Flask-Script==2.0.5
 Flask-SQLAlchemy==2.0
 Flask-SSLify==0.1.5
 Flask-WTF==0.11
-geopy==1.11.0
```
weronica: Isn't this used by the main site during production? `requirements/dev.txt` is (I think) meant to contain the requirements that are used only for development.

requirements/common.txt
```diff
@@ -1,8 +1,5 @@
 alembic==0.7.6
-async==0.6.2
```
weronica: Isn't this used by the main site during production?

app/main/views.py
armant: Because it's (going to be) information retrieval, namely querying all associations for a given resource. Doesn't make sense as a POST.

requirements/common.txt
armant: async? For what? I think this was added by Elizabeth by mistake, when she accidentally installed async node module (it was added in the merge request #26). It would've been here earlier if it was used in the production, right?

requirements/common.txt
armant: Oh is it? Ok, will add back. It was previously only in dev.txt, so I thought it was added by accident.

armant: Wait, isn't geopy only used for generating fake addresses? So both geopy and fake-factory should be either in dev or here, right?

requirements/common.txt
weronica: @ehamp is `async` needed?

requirements/common.txt
weronica: Ah my bad - I thought geopy was still being used to geocode addresses. If that's not the case, then we're good 👍

requirements/common.txt
ehamp: Yeah, @armant  I use async in the maps.js file

app/main/views.py
ehamp: Could someone explain why this is no longer a POST? Don't we need to post the name of the current pin to that route, in order to retrieve the data?

armant: No, as you can see, we're no longer passing the name, but passing the resource's ID instead. The convention for such things is to have a GET request and pass the ID in the URL. You can read more about when to use what type of request here: https://en.wikipedia.org/wiki/Representational_state_transfer (Relationship between URI and HTTP Methods)

requirements/common.txt
armant: Yep, but that async is not a Python package, but a Javascript script, which you've already added to assets/ .

armant: in requirements file you add the Python packages that this app depends on (i.e. dependencies). In other words, all those that you install with `pip install X`.

PR #31, opened 2016-04-28
Associations Sidebar

ehamp:
- Added a side bar that displays more information of any pin that is currently clicked

### DIFF
--- a/app/assets/scripts/vendor/map.js
+++ b/app/assets/scripts/vendor/map.js
@@ -122,11 +122,14 @@ function initMap() {
   mapViewButton.addEventListener('click', function() {
     $("#map").show();
     $("#list").hide();
+    $("#more-info").empty();
+    $("#sidebar").show();
   });
   listViewButton.addEventListener('click', function() {
     $("#map").hide();
     populateListDiv();
     $("#list").show();
+    $("#sidebar").hide();
   });
 }
 
--- a/app/assets/scripts/vendor/map.js
+++ b/app/assets/scripts/vendor/map.js
@@ -56,18 +56,6 @@ function initMap() {
     infowindow.open(map, marker);
   });
 
-  // Sets a listener on a radio button to change the filter type on Places
-  // Autocomplete.
-  function setupClickListener(id, types) {
-    var radioButton = document.getElementById(id);
-    radioButton.addEventListener('click', function() {
-      autocomplete.setTypes(types);
-    });
-  }
-
-  setupClickListener('changetype-all', []);
-  setupClickListener('changetype-address', ['address']);
-  setupClickListener('changetype-establishment', ['establishment']);
   var marker = new google.maps.Marker({
     map: map
   });
@@ -83,6 +71,8 @@ function initMap() {
    })
 
   function create_marker(resource){
+    //console.log(resource); resources dont have associations here
+    //why does .query.all() not return the associations?
     var markerToAdd = new google.maps.Marker({
        map: map
     });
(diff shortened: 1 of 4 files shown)


armant:
r @nnarang7

nnarang7:
lgtm!

nnarang7:
awesome work! looks good to me!

app/assets/scripts/vendor/map.js
```diff
@@ -83,6 +71,8 @@ function initMap() {
    })
 
   function create_marker(resource){
+    //console.log(resource); resources dont have associations here
```
armant: nit: delete the comment

app/assets/scripts/vendor/map.js
```diff
@@ -111,8 +99,16 @@ function initMap() {
             content: contentDiv
           });
           infowindow.open(map, markerToAdd);
+          google.maps.event.addListener(infowindow,'closeclick',function(){
```
armant: nit: pls add space after infowindow and 'closeclick'

app/assets/scripts/vendor/map.js
```diff
@@ -111,8 +99,16 @@ function initMap() {
...
+            $("#more-info").empty();
+          });
           $.get('get-associations/' + resource.id).done(function(associations) {
-            // TODO: write logic to display resource associations.
+            var associationObject = JSON.parse(associations);
+            $("#more-info" ).empty();
+            for(var key in associationObject){
```
armant: nit: space after for and )

app/assets/scripts/vendor/map.js
```diff
@@ -111,8 +99,16 @@ function initMap() {
...
+              var value = associationObject[key];
+              $("#more-info").append("<p>" + key+ ": " + value + "</p>");
```
armant: please use more "jQuery-style" way of creating the HTML elements, such as:
var p = document.createElement('p');
p.append(key, ': ', value);
$("#more-info").append(p);

Also, please bold the names. Same way:
var bolded = document.createElement('strong');

app/assets/styles/vendor/map.css
```diff
@@ -69,4 +66,17 @@ table {
   font-family: Roboto;
   font-size: 13px;
   font-weight: 300;
-} 
\ No newline at end of file
+}
+
+#sidebar {
```
armant: one of these two #sidebar-s is redundant? Also, some of the values are conflicting, make sure to leave the right ones :)

app/main/views.py
```diff
@@ -11,13 +11,18 @@ def index():
 
 @main.route('/get-resources')
 def get_resources():
-    resources = Resource.query.all()
+    resources = Resource.query.all()
```
armant: nit: remove trailing whitespace please :) (if you're using Sublime, there is an extension that takes care of it for you)

app/main/views.py
```diff
@@ -11,13 +11,18 @@ def index():
...
+    resources = Resource.query.all() 
     resources_as_dicts = Resource.get_resources_as_dicts(resources)
     return json.dumps(resources_as_dicts)
 
 
 @main.route('/get-associations/<int:resource_id>')
 def get_associations(resource_id):
-    # TODO: write logic to retrieve and return associations for resource.
-    associations = []
+    pin_info = Resource.query.filter_by(id=resource_id).first()
```
armant: pls use `resource = Resource.query.get(resource_id)` instead; a slightly more "canonic" way. It's better to call it a resource, because for the purposes of back-end, you are dealing with a resource object, and the fact that it'll be used for a pin in the front end is not very relevant here.

app/main/views.py
```diff
@@ -11,13 +11,18 @@ def index():
...
+    associations = {}
+    for pin in pin_info.text_descriptors:
```
armant: `for descriptor in resource. text_descriptors`

app/main/views.py
```diff
@@ -11,13 +11,18 @@ def index():
...
+        associations[pin.descriptor.name] = pin.text
+    for pin in pin_info.option_descriptors:
```
armant: `for descriptor in resource. option_descriptors `

app/main/views.py
armant: Before starting to iterate, add a check to make sure that a resource with such ID is found. Do it this way:
`if resource is None:
    return json.dumps(associations)`
i.e. if such resource doesn't exist, just return an empty object and exit.

PR #32, opened 2016-04-28
Add static pages support.

armant:


armant:
r @nnarang7 @dzhang55 
The relevant files are:
```
app/models/editable.py
app/templates/macros/page_macros.html
app/templates/main/about.html
app/templates/main/contact.html
app/main/views.py
app/models/__init__.py
app/templates/macros/nav_macros.html
requirements/common.txt
```

PR #33, opened 2016-04-29
CSV import

weronica:
- Create resources from a CSV file
- Existing descriptor schemata are deleted and replaced with those from CSV file. (This is _temporary_ -- in the future, the existing schema will be accommodated.)


weronica:
r @nnarang7 @dzhang55 @armant

PR #46, opened 2016-10-24
Update README.md

kyle-rosenbluth:
Add xcode-select note to readme


yoninachmany:
can u add to flask-base instead?

anniemeng:
I think adding to both would be helpful

PR #47, opened 2016-10-25
Text search for resource name

sanjayss34:


anniemeng:
Can you add screenshots for the changes and show the different workflows?
Searching -> Result on Map -> Switch to List
Searching -> Result on List -> Switch to Map
Removing Search Word -> Map View
Removing Search Word -> List View
etc.

sanjayss34:
beforemapsearch
aftermapsearch
aftermapremovesearch
Here are images showing the state before search, after search, and after removing the search keyword in the map view.

sanjayss34:
beforelistsearch
afterlistsearch
afterlistremovesearch
Here are images showing the state before search, after search, and after removing the search keyword in the list view.

app/main/views.py
```diff
@@ -19,6 +22,12 @@ def get_resources():
     resources_as_dicts = Resource.get_resources_as_dicts(resources)
     return json.dumps(resources_as_dicts)
 
+@main.route('/search-resources/<query_name>')
+def search_resources(query_name):
+    sys.stderr.write(query_name)
```
anniemeng: is this just for debugging? if so, can we remove this?

app/templates/main/index.html
```diff
@@ -10,9 +10,16 @@
...
+        <div>
+            <a href="{{ url_for('suggestion.suggest') }}">Suggest New Resource</a>
+            <form id="resources-form">
+                <input type="text" id="resources-input" class="controls"
+                       placeholder="Enter a resource keyword">
+            </form>
+        </div>
         <input id="pac-input" class="controls" type="text"
                placeholder="Enter a location">
+
```
anniemeng: remove extra space

app/main/views.py
```diff
@@ -6,6 +7,8 @@
 from .. import db
 from ..models import EditableHTML, Resource
 from . import main
+import sys
+import logging
```
anniemeng: remove unused

app/assets/scripts/vendor/map.js
```diff
@@ -142,6 +87,83 @@ function initMap() {
...
+  );
+}
+
+$(document).ready(function() {
+  $('#resources-form').submit(function(e) {
+    e.preventDefault();
+    var query = document.getElementById('resources-input').value;
+    $.get('/search-resources/'+query).done(function(resourcesString) {
+      for (var i=0; i < markers.length; i++) {
+        markers[i].setMap(null);
```
anniemeng: do we have to set these markers map to null if we clear them out in line 154?

sanjayss34:
Hi Annie, thanks, those changes have been made

app/assets/scripts/resources.js
```diff
@@ -0,0 +1,13 @@
+$(document).ready(function () {
+    // This is for search on the index page
+    $('#search-resources').submit(function(e) {
+        e.preventDefault();
+        var query = document.getElementById('search-resources-text').value;
+        console.log(query);
```
anniemeng: remove

app/assets/scripts/resources.js
```diff
@@ -0,0 +1,13 @@
...
+        var endpoint = '/single-resource/' + query;
+        if (query.length == 0) {
+            endpoint = '/single-resource/';
```
anniemeng: if query is an empty string wouldn't the endpoint be '/single-resource/' from line 7?

app/main/views.py
```diff
@@ -1,3 +1,4 @@
+from __future__ import print_function
```
anniemeng: remove

app/single_resource/views.py
```diff
@@ -14,8 +14,13 @@
 def index():
     """View resources in a list."""
     resources = Resource.query.all()
-    return render_template('single_resource/index.html', resources=resources)
+    return render_template('single_resource/index.html', resources=resources, query="")
```
anniemeng: instead of passing an empty query parameter I think it would be better to check if query is defined in the template itself?

PR #49, opened 2016-10-31
Flask base improvements

sandlerben:
@anniemeng I think we should merge almost all of these improvements in, except perhaps the python auto-formatting. What do you think?


app/templates/layouts/base.html
```diff
@@ -2,15 +2,39 @@
...
 <html>
-    {% include 'partials/_head.html' %}
-
+    <head>
+        {% include 'partials/_head.html' %}
+        {# Any templates that extend this template can set custom_head_tags to add scripts to their page #}
+        {% block custom_head_tags %}{% endblock %}
+    </head>
     <body>
+      {# Example dropdown menu setup. Uncomment lines to view
```
anniemeng: maybe we should remove these examples for our specific project?

anniemeng:
@sandlerben what do you think?

sandlerben:
yes, agreed. I'm very busy but will take a look later this week

sandlerben:
Adding all these changes isn't going to work, but I'll make another pr with the changes I think are important.

PR #53, opened 2016-11-02
Refactored index page and display map/list

sandlerben:


sandlerben:
Assigned to @stephanieyshi and @anniemeng for code review

anniemeng:
@sandlerben can you add screenshots for what you changed also?

sandlerben:
@anniemeng happy to this time, in the future let's definitely set-up heroku review apps to automate this

sandlerben:
image

app/assets/styles/map.scss
```diff
@@ -0,0 +1,29 @@
+#map {
+  height: 100%;
+  width: 100%;
+}
+
+#list {
+  height: 95%;
```
anniemeng: why 95%?

sandlerben: honestly, this was a bit of a hack. here's why.

the entire page has height 100%. within the page, there are two main elements which take up vertical space
- the header
- the map/list

if the map/list were height 100%, they would go over the page. so 95% accounts for the fact that the header takes up some room.

i'm totally open to doing this a better way though! any ideas?

anniemeng: Hm that is pretty hacky. Usually it'd be good to define an outer div encapsulating both elements but since I'm guessing you're talking about the header, which I'm assuming is the nav, we can't do that here. I think we can use this for now and I'll look more into how to structure the nav - there should be some set up to have nav not interfere with rest of page.

PR #56, opened 2016-11-07
Add Heroku generated app.json

anniemeng:


sandlerben:
I would rename the app

PR #60, opened 2016-11-12
Show all markers on list

sandlerben:


sandlerben:
@anniemeng can you look into why the deploy failed? also consider adding us all to the heroku app

anniemeng:
It's because it needs the updated config parameters from flask base which we're merging in on the PR you had of flask-base changes

PR #61, opened 2016-11-13
Refactor frontend view + resource expansion

anniemeng:
Break displaying map/list view into helper functions

Refactor how markers are displayed on the map, show all markers on the map
Add new circular location marker for location search
Add resource expansion page replacing the map when clicking a resource

TODO:
- Change how the resource expansion page elements are generated (currently a lot of manual injection and won't be feasible for ratings and review functionality)


app/assets/scripts/map.js
```diff
@@ -28,13 +39,9 @@ function initMap() {
...
     }
-    // If the place has a geometry, then present it on a map.
-    if (place.geometry.viewport) {
-      map.fitBounds(place.geometry.viewport);
-    } else {
-      map.setCenter(place.geometry.location);
-      map.setZoom(17);
-    }
+    map.setCenter(place.geometry.location);
+    map.setZoom(17);
```
sandlerben: Make a note that the map is re-centered/re-zoomed below

app/assets/scripts/map.js
```diff
@@ -46,38 +53,64 @@ function initMap() {
           place.address_components[2].short_name) || '')
       ].join(' ');
     }
-    marker.setLabel(address);
+
+    // create custom location marker
+    var marker = new google.maps.Marker({
+      map: map,
+      position: place.geometry.location,
+      anchorPoint: new google.maps.Point(0, -29),
```
sandlerben: what's this?

app/assets/scripts/map.js
```diff
@@ -46,38 +53,64 @@ function initMap() {
...
     var br = document.createElement('br');
     $(placeDiv).append(place.name, br, address);
     infowindow.setContent(placeDiv);
     infowindow.open(map, marker);
   });
+}
 
-  var marker = new google.maps.Marker({
-    map: map
+function initResourceSearch() {
```
sandlerben: I would document any new functions (and perhaps old ones too!)

app/assets/scripts/map.js
```diff
@@ -46,38 +53,64 @@ function initMap() {
...
+  $('#resources-form').submit(function(e) {
+    e.preventDefault();
```
sandlerben: What is e?

app/assets/scripts/map.js
```diff
@@ -46,38 +53,64 @@ function initMap() {
...
+    var query = $('#resources-input').val();
+    var endpoint = '/search-resources/'+query;
+    if (query.length == 0) {
+        endpoint = '/get-resources';
+    }
+    $.get(endpoint).done(function(resourcesString) {
+      for (var i=0; i < markers.length; i++) {
```
sandlerben: be consistent about spaces e.g. `i=0` or `i = 0`

app/assets/scripts/map.js
```diff
@@ -97,37 +130,62 @@ function create_marker(resource){
...
         var contentDiv = document.createElement('div');
-        var strong = document.createElement('strong');
+        var resName = document.createElement('strong');
+        $(resName).html(resource.name);
         var br = document.createElement('br');
-        $(strong).html(resource.name);
-        $(contentDiv).append(strong, br, resource.address);
+
+        // information pane to replace map
+        // TODO: refactor how the page elements are generated
```
sandlerben: There has to be a better way than this right? I hate the way we are creating html. Is there some light-weight client-side framework for generating html in js?

app/assets/scripts/map.js
anniemeng: this was from before, i can try removing it and see what happens

app/assets/scripts/map.js
anniemeng: The event e - this was added by sanjay (git thinks i added a lot of these things when I just move their location in the file). This is to prevent the browser from jumping url

app/assets/scripts/map.js
anniemeng: Yeah with my TODO refactor I'm going to create a separate task for this generating of info pages refactor since I think there'll be a lot there in and of itself. I'm currently looking into Mustache, Handlebars etc. as potentially useful but still have to do more research. If there isn't anything particularly useful then I'm hoping to have the corresponding Jinja file contain most of div components and just have this function populate the text content as an alternative instead of having both text and lot of div elements.

app/assets/scripts/map.js
anniemeng: New change: added note to method documentation

PR #70, opened 2016-11-20
Suggestions Views 

OBrandon:
Created a new resource on the top of the page. Added a table at the top of the page that contains the information about the resource for creating a new resource and editing a resource. When submitting a form it saves and sends an email to the admin. We fixed the viewing of what the admin sees when they go to suggested resources. Displays the suggested resource at the top of the new resource form.


app/templates/suggestion/create.html
```diff
@@ -0,0 +1,39 @@
+{% extends 'layouts/base.html' %}
+
+{% block content %}
+    <div class="ui text container">
+        <h2 class="ui dividing header">Add Resource</h2>
+        <div class="ui text container">
+			<table class="ui unstackable celled table">
```
anniemeng: Back indent everything within the `<div class="ui text container">` so it's only one indent level in.
In general for elements are "children" of the outer element, it should be one indent level in, and so if you have a chain of parent child parent child then it's a cascading effect

```
<div class="ui text container">
       <table class="ui unstackable celled table">
              <thead>     
```
etc.

Look how `app/templates/suggestion/edit.html` was done

PR #73, opened 2016-11-21
Styling for About and Contact

OBrandon:
Allowed you to exit out of the editor when you click outside of the editor box. Styling on the box fixed padding,margins and other things. Added a description that the user will see right under the header for about and contact us pages.

### DIFF
--- a/app/assets/styles/admin.scss
+++ b/app/assets/styles/admin.scss
@@ -1,15 +1,15 @@
 // About page 
 .updated_alert, .ui.button.cancel-edit, .ui.button.end-edit {
-	display: none;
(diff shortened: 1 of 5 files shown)


app/assets/styles/admin.scss
```diff
@@ -0,0 +1,15 @@
+// About page 
+.updated_alert, .ui.button.cancel-edit, .ui.button.end-edit {
+	display: none;
```
anniemeng: In general for scss, delete one level of indent than it is here for the properties
(similar indentation to how the corresponding css file is below)

PR #74, opened 2016-11-21
Added overlapping marker spiderfier and refactored in process

sandlerben:
- Markers which overlap now separate
- Modified resource add fake function to randomly create resources at same location
- Removed part of marker click listener which zoomed and re-centered map. The default behavior is *much* better.


app/assets/scripts/map.js
```diff
@@ -7,6 +7,64 @@ var map;
 var markers = [];
 var infowindow;
 
+// Click listener for a marker.
```
sandlerben: Note: This is copied near-verbatim from the previous click function.

sandlerben:
r @anniemeng @riyer15 @stephanieyshi

app/templates/descriptor/index.html
```diff
@@ -1,5 +1,6 @@
 {% extends 'layouts/base.html' %}
 
+<!-- TODO: This shouldn't be necessary. Remove if possible. -->
```
anniemeng: will remove in cleanup diff - thanks for this!

app/assets/scripts/map.js
anniemeng: the only thing you removed is what you noted right? the zoom/re-center stuff?

app/assets/scripts/map.js
```diff
@@ -4,9 +4,68 @@
 // <script src="https://maps.googleapis.com/maps/api/js?key=YOUR_API_KEY&libraries=places">
 
 var map;
+var oms;
 var markers = [];
 var infowindow;
 
+// Click listener for a marker.
+function markerListener(marker, event) {
```
anniemeng: is there a reason we don't use this event variable and pass null when we call the function?

app/assets/scripts/map.js
sandlerben: Yes

anniemeng: ok will incorporate that into my PR!

app/assets/scripts/map.js
anniemeng: Then can we replace the 
`moreLink.addEventListener('click', function() {` with 
`moreLink.addEventListener(event, function() {` instead? 
and then make the call `markerListener(markerToShow, 'click')` on line 251?
So that we don't hardcode the event and make use of the variable?

sandlerben: No, I don't think we should use a variable instead of a string literal for 'click'. I did modify 251.

PR #75, opened 2016-11-21
Refactor DOM element population to Handlebars templates

anniemeng:
Refactored the following DOM manipulation views:
1) Location search marker
2) General resource markers
3) Detailed resource information view
4) List view elements

to use Handlebars.js templates. The templates are defined in script tags in Jinja templates and then compiled and passed variable data in map.js

Use `{% raw %}{% endraw %}` to enclose Handlebars variables in Jinja file since they use the same notation as Jinja for variables ( `{{ var }}`) and don't want Jinja to parse these.


app/assets/scripts/map.js
```diff
@@ -14,7 +15,7 @@ var infowindow;
 function initMap() {
   map = new google.maps.Map(document.getElementById('map'), {
     center: {lat: 39.949, lng: -75.181}, // TODO(#52): Do not hardcode this.
-    zoom: 13
+    zoom: focusZoom,
```
sandlerben: This also probably shouldn't be hard-coded. I suppose it can be hard-coded here, but the map should dynamically update the zoom depending on the number of markers.

app/assets/scripts/map.js
```diff
@@ -74,10 +75,17 @@ function initLocationSearch(map) {
...
-    infowindow.setContent(placeDiv);
+
+    // Show marker info bubble for searched location
+    var locationMarkerTemplate = $("#locationmarker-template").html();
+    var compiledLocMarkerTemplate = Handlebars.compile(locationMarkerTemplate);
+    var context = {
+      name: place.name,
+      address: address,
+    };
+    var locationMarkerInfo = compiledLocMarkerTemplate(context);
```
sandlerben: This is awesome.

app/assets/scripts/map.js
```diff
@@ -74,10 +75,17 @@ function initLocationSearch(map) {
         fillOpacity: 1,
       },
     });
-    var placeDiv = document.createElement('div');
-    var br = document.createElement('br');
-    $(placeDiv).append(place.name, br, address);
-    infowindow.setContent(placeDiv);
+
+    // Show marker info bubble for searched location
+    var locationMarkerTemplate = $("#locationmarker-template").html();
```
sandlerben: Could you choose slightly more descriptive names than `locationMarkerTemplate` and `markerTemplate`?

app/assets/scripts/map.js
```diff
@@ -150,59 +159,76 @@ function create_marker(resource){
...
+        }
+        infowindow = new google.maps.InfoWindow({
+          content: markerInfo
+        });
+        infowindow.open(map, markerToAdd);
+        map.setCenter(markerToAdd.getPosition());
+        map.setZoom(focusZoom);
+
+        // Marker to detailed resource information view
+        $(".more-info").click(function() {
```
sandlerben: I am not super well-versed in javascript style, but perhaps a large anonymous function within another function would be better off defined outside of the function?

app/assets/scripts/map.js
```diff
@@ -150,59 +159,76 @@ function create_marker(resource){
...
           $.get('get-associations/' + resource.id).done(function(associations) {
             $("#map").hide();
             $("#resource-info").empty();
             $("#resource-info").show();
 
-            var backButton = document.createElement('button');
-            backButton.innerHTML = 'Back';
-            backButton.setAttribute('class', 'ui button');
-            backButton.addEventListener('click', function() {
+            var associationObject = JSON.parse(associations);
```
sandlerben: What does this do? Please document

app/assets/scripts/map.js
```diff
@@ -150,59 +159,76 @@ function create_marker(resource){
...
-              var value = associationObject[key];
-              $(descriptor).append(name, ': ', value);
-              $("#resource-info").append(descriptor);
-            }
-            var a = document.createElement('a');
-            $(a).attr('href', 'suggestion/' + resource.id);
-            $(a).html('Suggest an edit for this resource');
-            $("#resource-info").append(a);
+            // Map for single resource on detailed resource info page
+            var singleMap = new google.maps.Map(
```
sandlerben: Nit: consider renaming `singleMap` to `singleResourceMap`

app/assets/scripts/map.js
```diff
@@ -150,59 +159,76 @@ function create_marker(resource){
...
+              {
+                center: markerToAdd.getPosition(),
+                zoom: focusZoom,
+              }
+            );
+            var singleMarker = new google.maps.Marker({
+              position: markerToAdd.getPosition(),
+              map: singleMap,
+            });
           }).fail(function() {});
```
sandlerben: Not that you added this, but couldn't we go without `.fail`?

app/assets/scripts/map.js
```diff
@@ -218,34 +244,27 @@ function populateListDiv() {
...
+  var compiledListTemplate = Handlebars.compile(listTemplate);
+  var context = {
+    resource: listResources,
+  };
+  var listView = compiledListTemplate(context);
+  $("#list").html(listView);
 
-    listElement.addEventListener('click', function() {
-      google.maps.event.trigger(markerToShow, 'click');
+  // Can only add handlers to elements in template after compilation
```
sandlerben: What will the ui be when a new list of resources is rendered?

app/assets/styles/map.scss
```diff
@@ -8,8 +8,23 @@ $font-choice: Roboto;
 
 #list {
   height: 95%;
+  margin-top: 0px;
   overflow-y: auto;
-  padding: 10px;
+  padding: 0 0.5em 0 0.5em;
```
sandlerben: For my own education, why is `em` better than `px`?

sandlerben:
This is awesome and much appreciated! Left a few comments.

app/templates/main/index.html
```diff
@@ -119,6 +119,84 @@
       </div>
     </div>
 
+    <!-- HANDLEBARS.JS TEMPLATES COMPILED IN MAP.JS -->
+    <!-- Populate location search marker info-->
+    <script id="locationmarker-template" type="text/x-handlebars-template">
+      <div class="loc-marker-info">
+          <h3 class="ui header">
+              {% raw %}{{ name }}{% endraw %}
```
sandlerben: Does `{{{ name }}}` do the same thing?

app/assets/scripts/map.js
anniemeng: Yeah I'm thinking of just having this default value? and otherwise I'll remove the focus and zoom like you did in your PR and have that automatically adjust?

app/assets/scripts/map.js
anniemeng: yes, good point!

app/assets/scripts/map.js
anniemeng: yes good point 👍

app/assets/scripts/map.js
anniemeng: `JSON.parse` turns the JSON returned from the get request on line 184 into an JS object - I feel like the documentation is just what the method call itself does so is it necessary?

app/assets/styles/map.scss
anniemeng: `em` is a relative measure according to the font size and `px` is absolute measure so `em` makes it better for scaling according to the size of things around it instead of arbitrarily choosing `px` sizes that might only work for a certain font/screen size

app/assets/scripts/map.js
anniemeng: what do you mean? when a new search/filter is done to change the list or when the list is initially rendered? it would be the same in both cases since a new search just reruns this method which creates the UI outlined by the `#listview-template`

app/templates/main/index.html
anniemeng: No, you need the `{% raw %}{% endraw %}` to avoid it from being parsed by Jinja, Jinja still tries to parse `{{{ name }}}`

app/assets/scripts/map.js
sandlerben: I think I'm confused about the semantics of `associations`

app/assets/scripts/map.js
sandlerben: when a new search/filter changes the list. we may want to eventually add some animation/transition to the list div.

anniemeng: will make a separate task for this

app/assets/scripts/map.js
anniemeng: The naming of associations are from last sem, as descriptors are `OptionAssociation` and `TextAssociation` models

PR #80, opened 2016-11-28
Modify Admin search

anniemeng:
Make "All Resource" and "All Descriptor" search on Admin side by name of resource/descriptor only
TODO: Add filtering by descriptors to "All Resource" search

Add search button to "All Resource" search for future when need to also take into consideration filtering instead of just search query input - can add descriptor filtering functionality on Admin side here @kyle-rosenbluth @sanjayss34 

screen shot 2016-11-28 at 1 18 14 am

### DIFF
--- a/app/assets/scripts/descriptor.js
+++ b/app/assets/scripts/descriptor.js
@@ -1,23 +1,17 @@
 $(document).ready(function () {
 
     // This is for search on the index page
+    // Currently searches by descriptor name only
     $('#search-descriptors').keyup(function () {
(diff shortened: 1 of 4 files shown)


sanjayss34:
Looks good to me. Where would you like the search button and the dropdown to be when the filtering ui is added?

anniemeng:
hmm I don't have a set preference so you can play around with it and we can see? Maybe in similar format to user side with this stacked layout?

search input
descriptor filters
search button

PR #83, opened 2016-11-28
Modify User search

anniemeng:
Change order of search: put location on top and resource search on bottom, separated by a divider

Add search button functionality instead of submit form for resource input search to help with filtering @sanjayss34 @kyle-rosenbluth 

screen shot 2016-11-28 at 2 48 04 am


anniemeng:
Updated with @kyle-rosenbluth's ideas
screen shot 2016-11-28 at 6 50 48 pm

sanjayss34:
Do you think the "jump to location" should have a button rather than just being triggered by pressing enter?

anniemeng:
ooo that's an interesting idea! i'll add one and we can see how it looks

anniemeng:
So i tried adding another button for location and it ended up looking really cluttered. @riyer15 suggested something like this?
screen shot 2016-12-02 at 1 55 25 pm

sanjayss34:
Okay, I can see why it might be cluttered. As to the photo you posted, I don't see the difference between it and the previous screenshot except for the color changes. I suppose it should be clear enough to most people to press enter.

anniemeng:
Oh sorry! So the functionality with location search if you play around with it is that it autocompletes addresses and so if you click on one of the addresses that pops up as you type or enter a new valid address, it automatically triggers a place changed event which automatically will jump to location without pressing enter!

sanjayss34:
Oh, I see. Nice. Okay, looks good to me then.

PR #87, opened 2016-11-30
Map/List Grid Resizing

anniemeng:
When page loads and on window size change, resize the map/list grid to take up the whole viewable screen (adjust height of map/list grid container so that the height won't extend beyond the current viewport so user does not have to scroll to see whole container)

Refactor SCSS file to split styles into different documented parts

Add more specific styling for the map, list and resource views


stephanieyshi:
Was this supposed to be fixed? (when you scroll to the bottom of the list of resources)
screen shot 2016-12-01 at 1 06 50 am

anniemeng:
@stephanieyshi yeah are you on the right branch with all the recent changes pulled and are there are JS or SCSS errors? Since the background of the left panel should also be changed. This is what I see on the branch rn:

screen shot 2016-12-01 at 1 10 27 pm

stephanieyshi:
Ok, not sure why it wasn't updating correctly.

Is it possible to make the search section float over the list, or is that something that was decided against?

anniemeng:
haven't decided yet! this PR is to resize the map/list grid so it stays completely on the screen so we can discuss those changes and implement them in another PR

PR #92, opened 2016-12-02
Admin/User Fixes

OBrandon:
Added 'Suggest Resource' to Nav on USER side only after 'Contact Us' and made sure doesn't appear on Admin

 For 'About' and 'Contact Us' page on USER side, the text border box and description don't show up they only show up for admin

 When Admin edits 'About' and 'Contact Us' text editor and clicks out of it, the blue active state still shows until the user 'Cancel's


app/templates/macros/nav_macros.html
```diff
@@ -27,9 +27,9 @@
 {% endmacro %}
 
 {% macro account_items(current_user) %}
-    <a href="{{ url_for('main.about') }}" class="item">About</a>
```
anniemeng: I think we can keep this About and Contact Us here and remove them from the if else block since they will be there regardless of if it is a user or admin

app/templates/macros/page_macros.html
```diff
@@ -16,6 +15,10 @@
         Cancel
        </button>
        <div class="ui positive message updated-alert">Updated successfully</div>
+       {% else %}
```
anniemeng: back indent else

app/templates/macros/page_macros.html
```diff
@@ -28,6 +31,9 @@
                 startupFocus: true,
                 autoGrow_onStartup: true
             });
+            editor.on('blur',function(e){
```
anniemeng: space after `,` and `(e)`
I would also make a comment about what this is for

app/templates/macros/page_macros.html
```diff
@@ -70,9 +76,9 @@
             cancelEdit();
         });
 
-        /*$("#" + editorIDName).focusout(function() {
-            cancelEdit();
-        });*/
+        $("#" + editorIDName).focusout(function() {
```
anniemeng: add comment about what this is for

app/templates/macros/nav_macros.html
```diff
@@ -27,8 +27,8 @@
...
-    <a href="{{ url_for('main.contact') }}" class="item">Contact Us</a>
+        <a href="{{ url_for('main.about') }}" class="item">About</a>
```
anniemeng: let's back indent these about and contact

app/templates/macros/page_macros.html
```diff
@@ -70,9 +78,10 @@
             cancelEdit();
         });
 
-        /*$("#" + editorIDName).focusout(function() {
-            cancelEdit();
-        });*/
+        //Manages the focus activity of the CKEditor
```
anniemeng: I think we want to comment something about how we want the blue focus border to appear even if we focus out to denote an edit state? since this can be confusing to others without context

PR #100, opened 2016-12-07
Styling

stephanieyshi:
Styling User View
- color changes
- font changes
- padding/margin changes
- back button (resource view)
- scrolling for resource view

screen shot 2016-12-06 at 9 54 07 pm
screen shot 2016-12-06 at 9 54 18 pm


app/assets/styles/map.scss
```diff
@@ -1,5 +1,10 @@
 $accent-color: #2C9FA3;
 $font-choice: Roboto;
+$font-size: 14px;
+
+.ui.menu {
```
anniemeng: in general when styling semantic ui components i would add an extra class for that specific element because if in the future we add more `ui menu` items elsewhere in the app we might want to style it differently

app/assets/styles/map.scss
```diff
@@ -82,18 +96,21 @@ $font-choice: Roboto;
   display: block;
 }
 
-#resources-form {
-  display: inline;
-  text-overflow: ellipsis;
+.ui.horizontal.divider {
```
anniemeng: add more specific class to element

app/assets/styles/map.scss
```diff
@@ -82,18 +96,21 @@ $font-choice: Roboto;
...
-  text-overflow: ellipsis;
+#results-divider {
+  margin-bottom: 0px;
+}
+
+#search-keyword {
+  margin: 0 0 5px 0;
+}
+
+.ui.segment {
```
anniemeng: add more specific class to element

app/assets/styles/map.scss
```diff
@@ -104,44 +121,132 @@ $font-choice: Roboto;
...
+  margin: 0.5em;
+  padding-bottom: 0.5em;
   padding-top: 0.5em;
 
   .sub.header {
+    float: left;
+    font-family: $font-choice;
     font-weight: normal;
+    margin-right: 10px;
+    width: 80%;
```
anniemeng: instead of width could we do some padding?

app/assets/styles/map.scss
```diff
@@ -104,44 +121,132 @@ $font-choice: Roboto;
...
 
   .ui.label {
+    background-color: #DDDDDD;
+    display: inline-block;
     margin-left: 0;
     margin-top: 0.5em;
   }
+
+  a, a:hover, a:active {
+    color: $accent-color!important;
```
anniemeng: is there a way to add more selectors to avoid using `!important`?

app/assets/styles/map.scss
```diff
@@ -104,44 +121,132 @@ $font-choice: Roboto;
...
   width: 100%;
 }
 
 // Resource marker
+.marker-info {
+  font-family: $font-choice;
+  padding: 0.25em;
+}
+
+.ui.header {
```
anniemeng: add more specific class

app/assets/styles/map.scss
```diff
@@ -104,44 +121,132 @@ $font-choice: Roboto;
...
+}
+
 .more-info {
-  color: blue;
+  color: $accent-color;
   cursor: pointer;
   text-decoration: underline;
+
+  :hover {
+    color: $accent-color!important;
```
anniemeng: more specificity to avoid `!important`

app/assets/styles/map.scss
```diff
@@ -104,44 +121,132 @@ $font-choice: Roboto;
...
+  }
+}
+
+.more-info:hover {
+  color: $accent-color!important;
```
anniemeng: more specificity to avoid `!important`

PR #104, opened 2016-12-10
CSV improvements; handle update and reset

sandlerben:


app/bulk_resource/views.py
```diff
@@ -92,10 +93,95 @@ def upload_data():
...
+            resource.longitude = g.latlng[1]
+
+        for i, cell in enumerate(row.csv_body_cells):
+            if i not in csv_container.required_column_indices():
+                descriptor_name = \
+                    csv_container.csv_header_row.csv_header_cells[i].data
+                descriptor = Descriptor.query.filter_by(
+                    name=descriptor_name
+                ).first()
+                if descriptor is None:
```
anniemeng: If we introduce a new descriptor, wouldn't we want to add that to the db? This does introduce a lot more complexity though and would like your thoughts on this

app/bulk_resource/views.py
```diff
@@ -92,10 +93,95 @@ def upload_data():
...
+
+                text_association = [td for td in resource.text_descriptors if td.descriptor_id == descriptor.id]
+                text_association = text_association[0] if text_association else None
+                option_association = [od for od in resource.option_descriptors if od.descriptor_id == descriptor.id]
+                option_association = option_association[0] if option_association else None
+
+                if text_association is not None:
+                    db.session.delete(text_association)
+                elif option_association is not None:
+                    db.session.delete(option_association)
```
anniemeng: Correct me if I'm wrong but I'm assuming what you're doing is first deleting the option associations for the resource and then creating new associations right? We now support multiple option associations so instead of deleting `option_association[0]` you would have to delete all

app/bulk_resource/views.py
```diff
@@ -92,10 +93,95 @@ def upload_data():
...
+                    db.session.add(new_association)
+
+    if len(errors) > 0:
+        db.session.rollback()
+        for error in errors:
+            flash(error, 'error')
+        abort(400)
+
+    db.session.commit()
+    return redirect(url_for('single_resource.index'))
```
anniemeng: i'm thinking that if we add new descriptors in this csv upload we would also want to go to same sort of workflow as we do for reset in terms of choosing the type of descriptor it is?

app/bulk_resource/views.py
sandlerben: hmm, I think there should be just two use-cases

(a) reset: you set the schema for the database (veronica did this)
(b) update: you can update specific rows in the database, but you can't add new descriptors

do you think we should allow new descriptors with (b)? if so I would make a github issue

anniemeng: hm my main concern with not being able to add new descriptor columns by CSV is that they can either:
1) only add it through app interface which seems more work for them to click through the app vs. entering on csv & also their csv would be inconsistent with the data in the app
2) run a reset, but in that case we destroy all resources and build them up again which will destroy things like ratings/reviews on these resources as well as suggestions for existing resources

thoughts? also include @riyer15

app/bulk_resource/views.py
sandlerben: is this good now? @kyle-rosenbluth can you take a look

app/bulk_resource/views.py
sandlerben: see comment above

app/bulk_resource/views.py
anniemeng: yeah though I don't know if text associations will ever also be multiple but this works

app/bulk_resource/views.py
sandlerben: Anyway, this is separate from this PR so I say we merge this and make a github issue.

anniemeng: yeah this can be something we also discuss more in depth with the client @riyer15 and we can make an issue

app/bulk_resource/views.py
kyle-rosenbluth: This looks correct to me!

PR #108, opened 2016-12-12
Mobile responsive map/list grid view

anniemeng:
For a window size x:
`x > twoColResizeBreakpoint` 
We have the size of the list and map columns be what it was originally

` singleColBreakpoint < x < twoColResizeBreakpoint`  
We still have two columns but make the list column a little larger to keep it from being squished

` singleColNoSpaceBreakpoint < x < singleColBreakpoint`  
- We switch to a single column where it will show only a list view or only a map view. We default to the list view
- We also add a nav element that allows switching between the list and the map view
- Clicking an element in the list hides the list and displays the map view
- In the map view, we have a map as well as a map footer that will replace the infowindow to display the marker info and a link back to the list view
- For a location search, we go to the map view to display the searched location

`x <  singleColNoSpaceBreakpoint `  
Everything stays the same as the above range (`singleColNoSpaceBreakpoint < x < singleColBreakpoint`)  except that now there is no white space around the grid (we use the whole screen width since for smaller screens we need to make the grid as big as possible)

### DIFF
--- a/app/assets/scripts/map.js
+++ b/app/assets/scripts/map.js
@@ -274,6 +274,11 @@ function initLocationSearch(map) {
 
     infowindow.setContent(locationMarkerInfo);
     infowindow.open(map, locationMarker);
+
+    // If in single column view, entering a location should go to the map view
+    if ($(window).width() <= singleColBreakpoint) {
+      listToMapSingleColumn();
+    }
   });
 
   // Delete location marker when deleting location query
--- a/app/assets/scripts/map.js
+++ b/app/assets/scripts/map.js
@@ -9,6 +9,7 @@ var markers = [];
 var infowindow;
 var focusZoom = 17;
 var locationMarker;
+var allResourceBounds;
 
 // Click listener for a marker.
 function markerListener(marker, event) {
@@ -349,6 +350,7 @@ function populateMarkers(resources) {
 
   map.fitBounds(bounds);
   map.setCenter(bounds.getCenter());
+  allResourceBounds = bounds;
 }
 
 /*
@@ -474,7 +476,6 @@ function makeResponsive() {
   } else if ($(window).width() > twoColResizeBreakpoint) {
     $('#left-column').removeClass().addClass('four wide column');
     $('#right-column').removeClass().addClass('twelve wide column');
-    doubleColumnResets();
   }
 }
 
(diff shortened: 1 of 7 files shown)


app/assets/scripts/map.js
```diff
@@ -369,11 +410,138 @@ function populateListDiv() {
...
+  $('#map-footer').hide();
+  $('#left-column').show();
+  $('#right-column').show();
+  $('.nav-mobile-switch').hide();
 
   var center = map.getCenter();
   // Need to call resize event on map or creates dead grey area on map
   google.maps.event.trigger(map, "resize");
   map.setCenter(center);
+  map.setZoom(14);
```
riyer15: when a page is resized within the bounds of the desktop view, it zooms in too much & the center is determined in a different way from the default. try testing with resources far away from each other

app/templates/macros/nav_macros.html
```diff
@@ -14,6 +14,15 @@
...
         {% set href = url_for(current_user.role.index + '.index') %}
         <a class="item" href="{{ href }}">{{ current_user.role.name }} Dashboard</a>
+    {% else %}
+        <a class="item nav-mobile-switch" id="nav-to-map">
+          <i class="world icon"></i>
+          <p>Show Map</p>
+        </a>
+        <a class="item nav-mobile-switch" id="nav-to-list">
+          <i class="list icon"></i>
+          <p>Show List</p>
```
riyer15: what is the functionality when the user is on neither the map nor the list? (the about/other static pages?) need both views hidden

app/assets/scripts/map.js
```diff
@@ -483,8 +484,13 @@ function makeResponsive() {
 function singleColumnResets() {
   $('#left-column').removeClass().addClass('sixteen wide column');
   $('#right-column').removeClass().addClass('sixteen wide column');
-  $('#right-column').hide();
-  setNavSwitching();
+
+  // switched from double to single
+  // don't want to hide if a resize within single view is triggered
+  if ($('#right-column').is(':visible') && $('#left-column').is(':visible')) {
```
anniemeng: this should fix the issue of resizing map while in single column view disappearing issue

app/assets/styles/app.scss
```diff
@@ -8,7 +8,7 @@ nav {
   margin-bottom: 40px !important;
 }
 
-.nav-mobile-switch {
+.ui.menu .item.nav-mobile-switch {
```
anniemeng: fixes issue of seeing the 'Show Map' and 'Show List' on non-map/list pages

PR #110, opened 2016-12-12
Add title/placeholder to required option descriptor

anniemeng:
screen shot 2016-12-11 at 11 45 43 pm


riyer15:
LGTM @anniemeng

sanjayss34:
It might be a good thing to do a check in the template for whether req_desc is None. I don't think that would happen in the current setup, but it'd be a fairly easy way to avoid a None-type error. Otherwise, looks good to me.

anniemeng:
good catch @sanjayss34! added that in

PR #111, opened 2016-12-12
Fix ui forms

kyle-rosenbluth:
Add option descriptor selection in search. Filter using OR within a option, but AND between options.


app/main/views.py
```diff
@@ -8,10 +8,18 @@
...
 from ..models import EditableHTML, Resource, Rating, Descriptor, OptionAssociation, RequiredOptionDescriptor
 from . import main
+from wtforms.fields import SelectMultipleField, TextAreaField
+from ..single_resource.forms import SingleResourceForm
+
 from datetime import datetime
 
 @main.route('/')
 def index():
+    options = Descriptor.query.all()
```
anniemeng: we should also filter out the required descriptor from these so that we don't display a dropdown for it twice

app/assets/scripts/map.js
```diff
@@ -263,6 +263,17 @@ function initResourceSearch() {
     for (var i = 0; i < requiredOptions.length; i++) {
       query += '&reqoption=' + requiredOptions[i];
     }
+    var optionalOptions = [];
+    $("#advanced-search select").each(function() {
+        optionalOptions.push($(this).val());
+    })
+    var optionalOptionNames = [];
```
anniemeng: when is `optionalOptionNames` used?

app/main/views.py
```diff
@@ -20,14 +28,22 @@ def index():
...
-    return render_template('main/index.html', req_options=req_options)
+    return render_template('main/index.html', options=options_dict, req_options=req_options)
 
 @main.route('/get-resources')
 def get_resources():
     resources = Resource.query.all()
     resources_as_dicts = Resource.get_resources_as_dicts(resources)
     return json.dumps(resources_as_dicts)
 
+# @main.route('/search-resources/<query_name>')
```
anniemeng: do we need this?

app/main/views.py
```diff
@@ -41,7 +57,7 @@ def search_resources():
     req_opt_desc = Descriptor.query.filter_by(
         id=req_opt_desc.descriptor_id
     ).first()
-    resources = list(resource_pool)
+    resources = []
     if req_opt_desc is not None and len(req_options) > 0:
         resources = []
```
anniemeng: we can remove this line since we added it before the if statement

app/main/views.py
```diff
@@ -56,6 +72,36 @@ def search_resources():
                 if a.option in int_req_options:
                     resources.append(resource)
                     break
+    opt_options = request.args.getlist('optoption')
+    option_map = {}
+    for opt in opt_options:
+        if opt != "null":
```
anniemeng: can we do `if opt`

app/main/views.py
```diff
@@ -56,6 +72,36 @@ def search_resources():
...
+            for opt_val in option_val:
+                key_val = opt_val.split(':')
+                if key_val[0] in option_map:
+                    option_map[key_val[0]].append(key_val[1])
+                else:
+                    option_map[key_val[0]] = [key_val[1]]
+
+    descriptors = Descriptor.query.all()
+    new_pool = resource_pool
+    if len(req_options) > 0:
```
anniemeng: add comments

app/main/views.py
```diff
@@ -56,6 +72,36 @@ def search_resources():
                 if a.option in int_req_options:
                     resources.append(resource)
                     break
+    opt_options = request.args.getlist('optoption')
```
anniemeng: add comments

app/single_resource/views.py
```diff
@@ -171,11 +171,23 @@ def save_associations(resource, form, descriptors, resource_existed=True):
...
             association = None
             if resource_existed:
-                association = AssociationClass.query.filter_by(
+                associations = AssociationClass.query.filter_by(
                     resource_id=resource.id,
                     descriptor_id=descriptor.id
-                ).first()
+                )
+                for a in associations:
+                    print a
```
anniemeng: remove print

app/single_resource/views.py
```diff
@@ -171,11 +171,23 @@ def save_associations(resource, form, descriptors, resource_existed=True):
...
+                    db.session.delete(a)
+                    try:
+                        db.session.commit()
+                        flash('Resource deleted', 'form-success')
```
anniemeng: we're not deleting resources here right, just the association?

app/templates/main/index.html
```diff
@@ -1,119 +1,51 @@
...
+                    <div>{{ o }}</div>
+                    <select multiple="" class="ui fluid dropdown" placeholder="Select">
+                        {% for val in options[o] %}
+                                <option value="{{ o }}:{{ val }}">{{ val }}</option>
+                        {% endfor %}
+                    </select>
+                {% endfor %}
+            </div>
+
+            <div id="advanced-filters-showhide">
```
anniemeng: i would put this before your `#advanced-search` div so that when it is expanded, the hide link is at the top so the user doesn't have to scroll through all option descriptor filters in order to hide them

app/main/views.py
kyle-rosenbluth: Will that work? `opt` is actually a string holding the value "null"

app/single_resource/views.py
kyle-rosenbluth: Actually to clarify, that was my attempt at fixing the editing bug. It has now been commented out because if effectively does nothing.

app/main/views.py
anniemeng: can you try? what happens?

app/main/views.py
anniemeng: bump about commenting

app/templates/main/index.html
```diff
@@ -1,119 +1,50 @@
 {% extends 'layouts/base.html' %}
+{% import 'macros/form_macros.html' as f %}
```
anniemeng: do we use this anywhere?

app/main/views.py
```diff
@@ -16,11 +19,16 @@ def index():
     req_opt_desc = Descriptor.query.filter_by(
         id=req_opt_desc.descriptor_id
     ).first()
+    options = Descriptor.query.all()
+    options = [o for o in options if len(o.text_resources) == 0 and o.id is not req_opt_desc.descriptor_id]
```
anniemeng: we should first check if `req_opt_desc` is not None. If it is we should default it to some -1 value and then check if `o.id != -1` in that case

app/main/views.py
kyle-rosenbluth: Tried it out in the repl, if "null" evaluates to true.

app/templates/main/index.html
kyle-rosenbluth: I don't believe so. Good find.

app/main/views.py
kyle-rosenbluth: How would `o.id` ever be equal to 1?

app/main/views.py
kyle-rosenbluth: Comments added

app/main/views.py
anniemeng: How about if opt? No if "null"?

app/main/views.py
anniemeng: What do you mean by 1? I mean that sometimes a required option descriptor wont be set so we cant just call the descriptor_id as if it will always be there and have to set it to some dummy that will always evaluate to true like -1 since no id will be negative

anniemeng: essentially we wanna do something like
```
req_opt_id = -1
if req_opt_desc is not None:
   req_opt_id = req_opt_desc.descriptor_id
options = Descriptor.query.all()
options = [o for o in options if len(o.text_resources) == 0 and o.id != req_opt_desc.descriptor_id]
```
is how i'm thinking? but we should test

PR #112, opened 2016-12-12
Display required descriptors for each resource

anniemeng:
Display the required descriptors as labels for each resource on the list view and the detailed resource view

screen shot 2016-12-12 at 1 18 22 am
screen shot 2016-12-12 at 1 18 33 am


riyer15:
do we want to show the required option desc in the marker view as well?

anniemeng:
@riyer15 i'm not sure, i thought about that but also the marker window is already pretty populated and since there is the rating + the more info link + the possibility of having many required option desc but open to ideas?

sanjayss34:
Looks good to me.

PR #114, opened 2016-12-12
Fix bug of resource search view

anniemeng:
Bug: When viewing a detailed resource view for a resource and then entering a new search, the map doesn't fully show the resulting markers and the marker info window is extremely squished

Now: we always show the map when we search for a resource, and before the bounds are set so that the map can be bounded to show all the matching resources


riyer15:
the bug still exists in my version: 
screen shot 2016-12-12 at 2 34 13 am

riyer15:
sorry! closed by mistake

PR #121, opened 2017-01-03
Resource editing

riyer15:
Implemented the edit functionality for resource. Now handles multiple option descriptors, and also handles text and options

### DIFF
--- a/app/single_resource/views.py
+++ b/app/single_resource/views.py
@@ -162,8 +162,8 @@ def save_associations(resource, form, descriptors, resource_existed):
     handlers to the database."""
     #first delete all the associations for this resource if it already existed (to handle the "empty" case)
     if resource_existed:
-        options = OptionAssociation.query.filter_by(resource_id = resource.id).all()
-        texts = TextAssociation.query.filter_by(resource_id = resource.id).all()
+        options = OptionAssociation.query.filter_by(resource_id=resource.id).all()
+        texts = TextAssociation.query.filter_by(resource_id=resource.id).all()
         associations = options + texts
         for a in associations:
              db.session.delete(a)
@@ -171,6 +171,8 @@ def save_associations(resource, form, descriptors, resource_existed):
             db.session.commit()
         except IntegrityError:
             db.session.rollback()
+            flash('Error: failed to save edits. Please try again.',
+                  'form-error')
 
     for descriptor in descriptors:
         if descriptor.values:
--- a/app/single_resource/views.py
+++ b/app/single_resource/views.py
@@ -115,6 +115,7 @@ def edit(resource_id):
                 resource_id=resource_id,
                 descriptor_id=descriptor.id
             )
+            print option_associations
             if option_associations is not None:
                 default = [assoc.option for assoc in option_associations]
             setattr(SingleResourceForm,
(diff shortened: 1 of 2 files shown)


app/single_resource/views.py
```diff
@@ -115,6 +115,7 @@ def edit(resource_id):
                 resource_id=resource_id,
                 descriptor_id=descriptor.id
             )
+            print option_associations
```
anniemeng: delete print

app/single_resource/views.py
```diff
@@ -150,15 +151,32 @@ def edit(resource_id):
                   'form-error')
     # Field id is not needed for the form, hence omitted with [1:].
     for field_name in resource_field_names[1:]:
-        form[field_name].data = resource.__dict__[field_name]
+        if form[field_name]:
```
anniemeng: how come this is needed?

app/single_resource/views.py
```diff
@@ -150,15 +151,32 @@ def edit(resource_id):
...
                            resource_id=resource_id)
 
 
-def save_associations(resource, form, descriptors, resource_existed=True):
+def save_associations(resource, form, descriptors, resource_existed):
     """Save associations from the forms received by 'create' and 'edit' route
     handlers to the database."""
+    #first delete all the associations for this resource if it already existed (to handle the "empty" case)
+    if resource_existed:
+        options = OptionAssociation.query.filter_by(resource_id = resource.id).all()
```
anniemeng: no space around the `=`
also for the `texts` one too

app/single_resource/views.py
```diff
@@ -150,15 +151,32 @@ def edit(resource_id):
...
+        texts = TextAssociation.query.filter_by(resource_id = resource.id).all()
+        associations = options + texts
+        print "in this resource :)"
```
anniemeng: delete print

app/single_resource/views.py
```diff
@@ -150,15 +151,32 @@ def edit(resource_id):
...
+        for a in associations:
+             print a
```
anniemeng: delete print

app/single_resource/views.py
```diff
@@ -150,15 +151,32 @@ def edit(resource_id):
...
+             db.session.delete(a)
+             try:
+                 db.session.commit()
+                 print "association deleted"
```
anniemeng: delete print

app/single_resource/views.py
```diff
@@ -150,15 +151,32 @@ def edit(resource_id):
...
+             except IntegrityError:
+                 db.session.rollback()
+                 print "not deleting associations :/"
```
anniemeng: delete print

app/single_resource/views.py
anniemeng: instead of splitting to options and texts here, can you do AssociationClass query? since it doesn't seem we need the `options` and `texts` variables other than to delete?

app/single_resource/views.py
```diff
@@ -169,35 +187,15 @@ def save_associations(resource, form, descriptors, resource_existed=True):
...
-            else:
-                arguments = {'resource_id': resource.id,
-                             'descriptor_id': descriptor.id,
-                             keyword: value,
-                             'resource': resource,
-                             'descriptor': descriptor}
-                new_association = AssociationClass(**arguments)
-                db.session.add(new_association)
-
+            print value
```
anniemeng: remove prints

app/single_resource/views.py
```diff
@@ -150,15 +151,32 @@ def edit(resource_id):
...
+    #first delete all the associations for this resource if it already existed (to handle the "empty" case)
+    if resource_existed:
+        options = OptionAssociation.query.filter_by(resource_id = resource.id).all()
+        texts = TextAssociation.query.filter_by(resource_id = resource.id).all()
+        associations = options + texts
+        print "in this resource :)"
+        for a in associations:
+             print a
+             db.session.delete(a)
+             try:
```
anniemeng: i think we should put this `try` block of committing outside of the for loop? so we try once to commit all deletions instead of trying each one?

app/single_resource/views.py
riyer15: was getting an error that the field_name wasn't there for some of the resources? so i thought it was safe to do an if check

app/single_resource/views.py
riyer15: I'm deleting all the associations for the resource, regardless of if they're in the "descriptors" list passed into the function, because I wanted to handle the empty case (if you delete all the values in a field, it won't show up in the "descriptors" list). The AssociationClass can only be used for either text or option, but not for both (you have to assign it to the descriptor's type). Because I want to delete all of them, I have to do queries for both Text and Option and AssociationClass doesn't really exist/apply here.

app/single_resource/views.py
```diff
@@ -150,15 +150,28 @@ def edit(resource_id):
...
+    if resource_existed:
+        options = OptionAssociation.query.filter_by(resource_id = resource.id).all()
+        texts = TextAssociation.query.filter_by(resource_id = resource.id).all()
+        associations = options + texts
+        for a in associations:
+             db.session.delete(a)
+        try:
+            db.session.commit()
+        except IntegrityError:
+            db.session.rollback()
```
anniemeng: we should show some error if this happens and abort the operation i think

app/single_resource/views.py
anniemeng: ah okay makes sense! then let's just remove the space around the `=` on line 165 and 166

riyer15:
ok! should be fixed now

PR #123, opened 2017-01-11
CSV workflow revamp

anniemeng:
Redid the CSV upload workflow to accommodate both 'Update' and 'Reset' actions, as well as row by row processing of the CSV

**Reset:** completely wipe out all resources and descriptors already in the app. In the CSV workflow we don't reference any existing data and instead only care about the data in the CSV
**Update:**  add or update existing resources or descriptors. A resource in the CSV that has a name not matching any other resource in the app will be a new resource. A resource in the CSV with the same name as one already in the app will just have its address and/or descriptors updated. A descriptor column not in the app will be added as a new descriptor. A descriptor column already in the app by name will just have new possible values if it is an option descriptor. In the CSV workflow, we reference existing data.

## Client-side CSV processing:
screen shot 2017-01-10 at 10 45 32 pm

- Only accept .csv files to be uploaded
- Only activate the 'Upload' button if a file has been uploaded
- Create two columns to display file upload messages: one for success messages and one for error messages
- Before parsing, require that the user has the required columns of 'Name' and 'Address' in the CSV
- Trim whitespace from all CSV cells for processing
- Parse each CSV row by row using the `step` function of Papaparse instead of `complete` to handle potentially large files in the future
- Each row is going to be a dictionary of the header as keys (first row of the CSV - essentially the fields of 'Name', 'Address', diff descriptor names etc.) and cells in the row as values (ex: {'Name': 'Annie's shop', 'Address': '123 Chestnut Street'}). We store each of the data rows in an array to be sent to the server later
- While parsing each row, we check that it has the required fields of 'Name' and 'Address' and that they are not just empty, and that each row has all the fields, but still allow that the non-required field values (descriptors) can be empty
- While parsing each row, we check that we can geocode the input addresses. If we cannot geocode any addresses, we display exactly what addresses under the error messages and don't move onto the next CSV workflow step (all CSV resource addresses must be able to be geocoded in order to successfully upload the CSV)
- We will send requests to the server through a DeferredAjax object that is defined - this ensures that each Ajax request occurs sequentially meaning that the next Ajax request is only executed when the previous one finishes
- We first send a request with the header row (fields) to temporarily store what descriptors the CSV will be updating/adding. Then we send each of the rows of the CSV (resources) to temporarily store what resources the CSV will be updating/adding. These are all handled as DeferredAjax requests and the request is slightly different based on if it is part of an 'Update' or a 'Reset' request
- Finally, we send a DeferredAjax request to signal finished in order to move to next part of CSV processing

#### Storing the fields and rows from the client:
- Created a CSVStorage model which is the general object that will hold all the data for a single CSV and is linked to the user
- CSVRow is one row of the CSV and belongs to a CSVStorage. It stores the dictionary of fields to cell values for each row (essentially is each resource in the CSV). It will have a `resource_id` if it is linked to an existing resource (same name distinguishes if resources are the same)
- CSVDescriptor is one field of the CSV and belongs to a CSVStorage. It stores the values and type of this descriptor. It has a `descriptor_id` if it is already in the app (has same name)
- CSVDescriptorRemove is a field in the app that is not in the CSV: meaning that this descriptor should be removed. Currently it's use is not implemented and we need to talk to client about if this is something they should be able to do: deleting descriptors through CSV

#### Deciding descriptor types:
- Show all the newly added descriptors in a form that allows selection of if the descriptor is an 'Option' or 'Text'. We only allow this selection for if it is a descriptor not already in the app since we don't allow a user to change an existing descriptor's type (they have to delete and re-add it in order to do so). Thus, if there are any descriptors in the CSV that are already in the app, we just show their type for the user's information.

*Reset*:
screen shot 2017-01-10 at 11 08 28 pm

*Update:*
screen shot 2017-01-10 at 11 59 09 pm

#### Showing the 'option' descriptor parsed values:
- For each of the descriptors in the CSV that the user selected will be an option descriptor, we look through each row in the CSV and find what values were assigned for this field and display it to the user to verify. Each option descriptor is semicolon delimited in a CSV cell. If the user thinks something is wrong, we make them re-upload the CSV
- Removed the ability to add additional descriptor option values by input form - should only be done through CSV to ensure that the CSV state and app state are as closely aligned as possible, and being able to add additional values not in the CSV while in the CSV workflow is confusing
- If the descriptor is already in the app, we show them what option values are in the CSV and what option values are in the app.

*Reset*:
screen shot 2017-01-10 at 11 08 52 pm

*Update:*
screen shot 2017-01-10 at 11 59 42 pm

#### Selecting required option descriptor:
- From the specified option descriptors in the CSV, require the user to choose one of them as the required one (will be a primary filter when users want to find a resource)
- If we are doing an 'Update', also allow the user to keep the current required descriptor if there is one. The user can only choose required descriptors from the option descriptors in the CSV or the current required descriptor. They can't choose any option descriptors that are already in the app but not in the current CSV
- If there are no option descriptors at all, we don't show this page and continue to 'Save'

*Reset*:
screen shot 2017-01-10 at 11 35 54 pm

*Update:*
screen shot 2017-01-11 at 12 02 53 am

#### Validate required option descriptor:
- For the chosen required option descriptor, if any of the resources in the CSV are missing a value for it, we display those resources and force the user to add a value for them to enforce the required constraint
- If doing an 'Update', we also show the resources in the app that lack this required option descriptor and force the user to add a value for them

screen shot 2017-01-11 at 12 05 42 am

#### Save
- Here for a 'Reset', we delete all the existing Resource and Descriptor data in the app
- From the temporary resource and descriptor data stored as CSVRow and CSVDescriptor, we either update or create new Resource and Descriptor entries
- We also set the required option descriptor and create a descriptor association for the resources that lacked a value before for this required option descriptor

*TODO: styling*
screen shot 2017-01-10 at 11 56 05 pm

## NOTE
The git preview is a little wonky so I recommend just looking straight at the whole raw new file for:
`app/bulk_resource/views.py`
`app/templates/bulk_resource/upload.html`
since majority of the file content has changed

## TODO
- Styling
- Possibly removing descriptors from the app if not in the CSV for 'Update' depending on client


app/templates/bulk_resource/upload.html
```diff
@@ -18,160 +18,256 @@ <h2 class="ui header">
...
+      var fields;
+      $("#csv-file").parse({
+        config: {
+          delimiter: ",",
+          header: true,
+          skipEmptyLines: true,
+          step: function(row, parser) {
+            // Check CSV formatting before parsing first row
+            if (numRows == 0) {
+              fields = row.meta.fields.map(e => e.trim()).filter(e => e !== '');
```
abhisuri97: Careful here. older browsers may not support `=>` aka fat arrow function syntax.

abhisuri97: there are multiple instances, just pointing this out in the first instance

app/templates/bulk_resource/upload.html
```diff
@@ -18,160 +18,256 @@ <h2 class="ui header">
...
+                <div class="ui list" id="status-errors"></div>
+              </div>
             </div>
         </div>
     </div>
 
-    <script src="http://maps.google.com/maps/api/js?sensor=false" type="text/javascript"></script>
-    <script type="text/javascript">
-        $('.ui.radio.checkbox').checkbox();
+  <script type="text/javascript">
```
abhisuri97: It would be good style to put this in a separate file. it can be placed in a separate file as is I believe as long as it is loaded after `jquery`

app/templates/bulk_resource/upload.html
```diff
@@ -18,160 +18,256 @@ <h2 class="ui header">
...
+          "<div class='item'>" + errorList[i] + "</div>"
+        );
+      }
+    }
+
+    // Helper to submit the parsed CSV data to the server to go through the
+    // CSV workflow
+    function submitCsvData(numRows, rowObjects, fields) {
+      // Needed for CSRF protection.
+      $.ajaxSetup({
```
abhisuri97: you can also put the `@csrf_exempt` decorator on the server-side endpoint. either works.

app/templates/bulk_resource/upload.html
```diff
@@ -18,160 +18,256 @@ <h2 class="ui header">
...
-        </div>
-        <div class="actions">
-            <div class="ui right labeled button" id="close-button">
-                Got it!
+                <button class="ui disabled button" type="submit" id="upload-button">Upload</button>
+            </form>
+            <h3 class="ui header">Upload Progress</h3>
+            <div class="ui horizontal segments" id="progress-msgs">
+              <div class="ui segment">
+                <h5 class="ui header green">Success Messages</h5>
```
abhisuri97: Limit length of this div
 (also give an idea to the user that something is happening). Perhaps a progress bar might serve you better here?

app/bulk_resource/views.py
```diff
@@ -41,438 +41,593 @@
...
-                if i not in csv_container.required_column_indices():
-                    descriptor_name = \
-                        csv_container.csv_header_row.csv_header_cells[i].data
+        # Create/update rows and descriptor associations
+        for row in csv_storage.csv_rows:
+            if csv_storage.action == 'update' and row.resource_id:
+                resource = Resource.query.filter_by(
+                    id=row.resource_id
+                ).first()
+                if resource.address != row.data['Address']:
```
anniemeng: i'm adding that in next push! it's currently a TODO

app/bulk_resource/views.py
```diff
@@ -41,438 +41,593 @@
...
+
+    # Collect all option descriptors in the CSV to display as choices in the
+    # SelectField.
+    for desc in csv_storage.csv_descriptors:
+        if desc.descriptor_type == 'option' and desc.name != req_name:
+            descriptors.append(desc.name)
+
+    # No existing required option descriptor and no option descriptors in CSV,
+    # then no possible required option descriptor
+    if len(descriptors) == 0:
```
anniemeng: hm it's not enforced on app side but I'm not sure about if it should/should not be required? I feel like without one it is harder for the user to find resources? since without option descriptors, we would have no primary or secondary filters?

anniemeng:
- Added address geocoding check (temporarily switch between two Google Maps API Key - will discuss with the client about possible options - either various dummy google accounts or upgrading plan)
- Remove requirement for setting a required resource descriptor
- Various bugs

PR #124, opened 2017-01-14
Delete Option Functionality Streamlined

riyer15:
When deleting an option, the user is shown a list of the affected resources and links to their edit pages before confirming they want to delete that option.

### DIFF
--- a/app/descriptor/views.py
+++ b/app/descriptor/views.py
@@ -247,10 +247,8 @@ def remove_option_value(desc_id, option_index):
 
     form = FixAllResourceOptionValueForm()
 
-    # Delete the dynamic fields after the form is instantiated
     if form.validate_on_submit():
         for oa in option_assocs:
-            # Case for 'Remove this descriptor'
             db.session.delete(oa)
 
         if remove_value_from_db(descriptor, choice_names, old_value):
--- a/app/descriptor/views.py
+++ b/app/descriptor/views.py
@@ -2,7 +2,6 @@ from flask import abort, flash, render_template, redirect, url_for, request
 from flask.ext.login import login_required
 from sqlalchemy.exc import IntegrityError
 from wtforms.fields import SelectField
-from wtforms.fields import SelectMultipleField
 from flask_wtf.file import InputRequired
 
 from forms import (
@@ -106,9 +105,11 @@ def edit_name(desc_id):
         db.session.add(descriptor)
         try:
             db.session.commit()
+            flash('Name for descriptor {} successfully changed to {}.'.format(old_name, descriptor.name),
+                'form-success')
         except IntegrityError:
             db.session.rollback()
-            flash('Database error occurred, and the name was not changed. Please try again.', 'form-error')
+            flash('Database error occurred. Please try again.', 'form-error')
         return render_template('descriptor/manage_descriptor.html',
                                desc=descriptor, is_option=is_option)
     form.name.data = descriptor.name
@@ -244,7 +245,6 @@ def remove_option_value(desc_id, option_index):
         remove_value_from_db(descriptor, choice_names, old_value)
         return redirect(url_for('descriptor.descriptor_info', desc_id=desc_id))
 
-    # Create the select field for each resource.
     form = FixAllResourceOptionValueForm()
 
     # Delete the dynamic fields after the form is instantiated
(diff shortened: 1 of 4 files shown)


app/descriptor/views.py
```diff
@@ -243,48 +241,35 @@ def remove_option_value(desc_id, option_index):
...
-            if choice == -1:
-                db.session.delete(oa)
-            else:
-                oa.option = choice
-                db.session.add(oa)
-
-        # Index starting from 1 to skip 'Remove this descriptor'
-        if remove_value_from_db(descriptor, choice_names[1:], old_value):
+            db.session.delete(oa)
+            print oa
```
anniemeng: remove print

app/descriptor/views.py
```diff
@@ -243,48 +241,35 @@ def remove_option_value(desc_id, option_index):
...
+        remove_value_from_db(descriptor, choice_names, old_value)
         return redirect(url_for('descriptor.descriptor_info', desc_id=desc_id))
 
     # Create the select field for each resource.
-    for oa in option_assocs:
-        setattr(FixAllResourceOptionValueForm, oa.resource.name,
-                SelectField('', coerce=int, choices=choices))
     form = FixAllResourceOptionValueForm()
 
     # Delete the dynamic fields after the form is instantiated
```
anniemeng: remove comment as well

app/descriptor/views.py
```diff
@@ -243,48 +241,35 @@ def remove_option_value(desc_id, option_index):
...
-            else:
-                oa.option = choice
-                db.session.add(oa)
-
-        # Index starting from 1 to skip 'Remove this descriptor'
-        if remove_value_from_db(descriptor, choice_names[1:], old_value):
+            db.session.delete(oa)
+            print oa
+
+        if remove_value_from_db(descriptor, choice_names, old_value):
```
anniemeng: we should probably also show an error if this doesn't succeed

app/descriptor/views.py
```diff
@@ -243,48 +241,35 @@ def remove_option_value(desc_id, option_index):
...
-    choice_names = (['Remove this descriptor'] +
-                    descriptor.values[:removed_index] +
+    """Helper function to generate the option values for a SelectMultipleField."""
+    choice_names = (descriptor.values[:removed_index] +
                     descriptor.values[removed_index + 1:])
     choices = []
     for i in range(len(choice_names)):
-        choices.append((i - 1, choice_names[i]))
+        choices.append((i, choice_names[i]))
+    print choices
```
anniemeng: remove print

app/templates/descriptor/confirm_resources.html
```diff
@@ -42,16 +42,12 @@ <h3 class="ui red block header">
                         <thead>
```
anniemeng: in line 24 where we mention which option descriptor value, i think we should emphasize it more since it seems hard to read:
screen shot 2017-01-14 at 11 04 52 pm

maybe putting bold `<b></b>` tags around the option value name?

app/templates/descriptor/confirm_resources.html
```diff
@@ -42,16 +42,12 @@ <h3 class="ui red block header">
...
                         </thead>
                         <tbody>
-                        {% for i in range(option_assocs|length) %}
+                        {% for oa in option_assocs %}
                             <tr>
-                                <td>{{ option_assocs[i].resource.name }}</td>
-                                <td class="descriptor value">
-                                    {{ f.render_form_field(form[option_assocs[i].resource.name])}}
-                                </td>
+                                <td><a href="{{ url_for('single_resource.edit', resource_id=oa.resource['id']) }}">{{ oa.resource.name }}</a></td></td>
```
anniemeng: can we access the resource id directly from the option association? `oa.resource_id`?

app/templates/descriptor/manage_descriptor.html
```diff
@@ -66,6 +66,7 @@ <h3 class="ui header">Values</h3>
                 <tr>
                     <td colspan="2">{{ f.render_form_field(form.value) }}</td>
                     <td>{{ f.render_form_field(form.submit) }}<td>
```
anniemeng: you didn't add this but can you change the end tag to `</td>`

app/descriptor/views.py
```diff
@@ -2,6 +2,7 @@
 from flask.ext.login import login_required
 from sqlalchemy.exc import IntegrityError
 from wtforms.fields import SelectField
+from wtforms.fields import SelectMultipleField
```
anniemeng: do we use this anywhere?

app/descriptor/views.py
```diff
@@ -243,48 +241,35 @@ def remove_option_value(desc_id, option_index):
...
                            option_assocs=option_assocs, desc_id=desc_id,
                            desc=descriptor, option_index=option_index,
                            form=form)
 
-
 def generate_option_choices(descriptor, removed_index):
-    """Helper function to generate the option values for a SelectField."""
-    choice_names = (['Remove this descriptor'] +
-                    descriptor.values[:removed_index] +
+    """Helper function to generate the option values for a SelectMultipleField."""
```
anniemeng: we can reword this comment since i don't think we're using these values for any type of SelectField?

riyer15:
fixed for comments!! @anniemeng

app/descriptor/views.py
anniemeng: bump to delete comment

app/descriptor/views.py
```diff
@@ -243,48 +242,34 @@ def remove_option_value(desc_id, option_index):
...
     form = FixAllResourceOptionValueForm()
 
     # Delete the dynamic fields after the form is instantiated
-    for oa in option_assocs:
-        delattr(FixAllResourceOptionValueForm, oa.resource.name)
-
     if form.validate_on_submit():
         for oa in option_assocs:
-            choice = form[oa.resource.name].data
             # Case for 'Remove this descriptor'
```
anniemeng: also remove comment since there are no more cases

riyer15: the comments you're making are a different version of the file than i have locally! will re-push

riyer15:
changed @anniemeng

PR #125, opened 2017-01-14
fixed so descriptor edit name does not display error on success

riyer15:
Descriptor edit name functionality flashes fixed! previously showed errors on both success and failure, now only shows error on failure.

### DIFF
--- a/app/descriptor/views.py
+++ b/app/descriptor/views.py
@@ -93,7 +93,6 @@ def edit_name(desc_id):
     old_name = descriptor.name
     form = EditDescriptorNameForm()
     if form.validate_on_submit():
-        descriptor.name = form.name.data
         if Descriptor.query.filter(Descriptor.name == form.name.data).first() \
                 is not None:
             flash('Descriptor {} already exists.'.format(descriptor.name),
@@ -101,7 +100,8 @@ def edit_name(desc_id):
             return render_template('descriptor/manage_descriptor.html',
                                    desc=descriptor, form=form,
                                    is_option=is_option)
-
+            
+        descriptor.name = form.name.data
         db.session.add(descriptor)
         try:
             db.session.commit()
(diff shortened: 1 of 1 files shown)


app/descriptor/views.py
```diff
@@ -93,15 +93,15 @@ def edit_name(desc_id):
     old_name = descriptor.name
     form = EditDescriptorNameForm()
     if form.validate_on_submit():
-        descriptor.name = form.name.data
         if Descriptor.query.filter(Descriptor.name == form.name.data).first() \
                 is not None:
             flash('Descriptor {} already exists.'.format(descriptor.name),
```
anniemeng: since we don't change descriptor name to form name until after, we should print `form.name.data` instead of `descriptor.name`

app/descriptor/views.py
```diff
@@ -93,15 +93,15 @@ def edit_name(desc_id):
     old_name = descriptor.name
     form = EditDescriptorNameForm()
     if form.validate_on_submit():
-        descriptor.name = form.name.data
         if Descriptor.query.filter(Descriptor.name == form.name.data).first() \
```
anniemeng: right now we also show error if we change the name to be the same as it was before, so if we were in the 'Gender' descriptor and we try just changing name to 'Gender' (basically leaving it unchanged), we still show the error that the descriptor 'Gender' already exists. Should we handle this differently by either showing a warning that there is no change or just allow the change?

riyer15: I'll show a warning that there is no change, and we just won't do the change -- so then we don't have to make an uneccessary db change?

riyer15:
fixed for code review@ @anniemeng

app/descriptor/views.py
anniemeng: okay sounds good!

PR #128, opened 2017-01-14
Add reset search box

anniemeng:
Added a 'Reset' button to the search box that clears the search input and selected options in the dropdowns, and then shows all resources again

screen shot 2017-01-13 at 11 54 20 pm

### DIFF
--- a/app/assets/scripts/map.js
+++ b/app/assets/scripts/map.js
@@ -207,13 +207,11 @@ function initMap() {
   infowindow = new google.maps.InfoWindow();
   initLocationSearch(map);
   initResourceSearch();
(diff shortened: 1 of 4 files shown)


riyer15:
when I try to do the reset functionality, it doesn't work! not sure what's going on bc it seems to fail without an error

anniemeng:
@riyer15 hm that's strange - no console or python errors?
it works locally for me and also on the heroku (https://maps4all-pr-128.herokuapp.com)

riyer15:
yeah that's erally strange -- when i test it on heroku it makes sense & works. the code also looks good! lgtm

PR #131, opened 2017-01-19
Advanced CKEditor functionality

anniemeng:
Created new CKEditor package:
- Full version with more advanced options
- Standard editor frame instead of floating inline

screen shot 2017-01-19 at 6 11 07 pm

Other:
- Allow updated successfully message to be closed
screen shot 2017-01-19 at 6 23 35 pm


riyer15:
lgtm!!

PR #137, opened 2017-01-23
Add instructions

riyer15:
added more detailed instructions for inserting links and phone numbers in static pages

### DIFF
--- a/app/templates/macros/nav_macros.html
+++ b/app/templates/macros/nav_macros.html
@@ -73,6 +73,45 @@
     {% endif %}
 {% endmacro %}
 
+
+{% macro mobile_account_items(current_user) %}
+    <a href="{{ url_for('main.about') }}" class="item">About</a>
+    <a href="{{ url_for('contact.index') }}" class="item">Contact Us</a>
+    {% if current_user.is_authenticated() %}
+        <div class="ui simple dropdown item">
+            Manage Resources
+            <i class="dropdown icon"></i>
+            <div class="menu">
+                <h3 class="ui header">Resources</h3>
+                {{ dropdown_choice('All Resources', 'single_resource.index',
+                                    icon='book icon') }}
+                {{ dropdown_choice('Add New Resource', 'single_resource.create',
+                                    icon='add circle icon') }}
+                {{ dropdown_choice('Suggested Resources', 'suggestion.index',
+                                    icon='checkmark box icon', id='suggested-resources') }}
+                <div class="ui section divider"></div>
+                <h3 class="ui header">Descriptors</h3>
+                {{ dropdown_choice('All Descriptors', 'descriptor.index',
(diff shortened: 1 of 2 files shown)


riyer15:
fixed comments!!

app/templates/macros/nav_macros.html
```diff
@@ -122,3 +122,57 @@ <h3 class="ui header">Bulk</h3>
         {{ desktop_nav(current_user, endpoints=endpoints, count=count) }}
     </nav>
 {% endmacro %}
+) }}" class="item">Log out</a>
```
anniemeng: did you make additional changes to this file?
i think maybe something went wrong on a merge?

riyer15:
ok i think the nav is fixed again?

app/templates/macros/nav_macros.html
```diff
@@ -84,10 +123,8 @@ <h3 class="ui header">Bulk</h3>
 
         {# The menu items which will be shown when open-nav is clicked #}
         <div class="ui fixed vertical fluid menu">
-            {% if endpoints %}
```
anniemeng: i think we should keep this `endpoints` check

app/templates/macros/nav_macros.html
```diff
@@ -73,6 +73,45 @@ <h3 class="ui header">Bulk</h3>
     {% endif %}
 {% endmacro %}
 
+
+{% macro mobile_account_items(current_user) %}
+    <a href="{{ url_for('main.about') }}" class="item">About</a>
+    <a href="{{ url_for('contact.index') }}" class="item">Contact Us</a>
+    {% if current_user.is_authenticated() %}
+        <div class="ui simple dropdown item">
```
anniemeng: for mobile we should remove this dropdown right

app/templates/macros/nav_macros.html
riyer15: hmm yeah i want to just remove this file from the commit tbh

riyer15: i thought it was the right version ! i guess not hm

app/templates/macros/nav_macros.html
riyer15: we're not mobile responsive for admin so not really, but i can make that change and commit it if we want

app/templates/macros/nav_macros.html
riyer15: OH wait i figured it out: the fixed nav is only in the m4a-jlc-sp2 branch because I only fixed the dropdown for the user side (we don't have a dropdown for static pages in the generalized product) --this version of the file is messed up, but i reset it to the master version and i think it should be good now.

riyer15:
ok! i reset the nav_macros file so it isn't in this PR, just the page_macros changes. hope this is better!

PR #141, opened 2017-01-24
Review model definitions

anniemeng:
- Increase length of string fields in models (around 500 char)
- Delete suggestions attached to resources (and not suggestions not linked to an existing resource) and ratings when reseting app through CSV upload (since these are attached to specific resources)
- Fix setting of required option descriptor (bug that assumed always a required option descriptor to be set, but now allow for not having a required option descriptor) in CSV workflow
- Add on delete cascade for foreign keys
- Add commenting to models and some code style fixes

### DIFF
--- a/app/bulk_resource/views.py
+++ b/app/bulk_resource/views.py
@@ -29,9 +29,11 @@ from ..models import (
     GeocoderCache,
     Descriptor,
     OptionAssociation,
+    Rating,
     Resource,
     RequiredOptionDescriptor,
     RequiredOptionDescriptorConstructor,
+    Suggestion,
     TextAssociation
 )
 from forms import (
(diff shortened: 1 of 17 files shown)


app/bulk_resource/views.py
```diff
@@ -599,6 +601,8 @@ def save_csv():
         if csv_storage.action == 'reset':
             OptionAssociation.query.delete()
             TextAssociation.query.delete()
+            Suggestion.query.delete()
```
riyer15: is there a reason we delete all the suggestions? could imagine a use case where we reset the DB but still want to see all the suggestions (not resource-specific ones but the general suggestions?)

anniemeng: oh right, good catch!

PR #142, opened 2017-01-25
CSV workflow styling and instructions

anniemeng:
- Fixed up CSV workflow instructions for each step
- Added styling to last page of workflow: 'Save' by adding the Semantic UI three button styling
- More consistent styling in displaying of instructions
- Remove unused templates used in old CSV workflow


app/templates/bulk_resource/review_desc_options.html
```diff
@@ -31,15 +31,19 @@
...
-            For each of the columns that you marked as options, we have
-            identified the options already present in your CSV file. If
-            any of the existing options are incorrect, please edit the original
-            CSV file and re-upload it. We will store the options in alphabetical
-            order.
+          <h4 class="ui header">
+            Review Option Descriptor Option Values
+            <div class="sub header">
+              For each of the column headers that you marked as option descriptors,
+              we have identified the option values present in your CSV file. If
```
riyer15: can we write that the CSV adds to these options? at first read, the language implies that we're overwriting them with CSV. Also write that you can add option values in-app.

riyer15:
screen shot 2017-01-25 at 11 20 02 am
update didn't recognize this duplicate name -- do you know what could have happened?

riyer15:
i just tried doing it again and update was correct -- maybe it was an error with Heroku having its own data, i'm not totally sure.

app/templates/bulk_resource/save.html
```diff
@@ -28,10 +28,21 @@
...
-            {{ f.render_form(form) }}
+            <h2 class="ui header">
+                CSV Upload Workflow
+            </h2>
+        </div>
+        <div class="twelve wide column">
+          <h4 class="ui header">
+            Save Changes
+            <div class="sub header">
+              It may take awhile to save your changes.
```
riyer15: add a space between "a" and "while"

riyer15:
screen shot 2017-01-25 at 11 26 03 am
got this error after clicking "back" at the "Save" step

anniemeng:
@riyer15 can you pull master and try again? i think it was from a bug solved in another PR

riyer15:
@anniemeng i found the issue on the heroku version -- will try running locally

riyer15:
getting this issue locally (in logs): File "/Users/macbookpro/Documents/hack4impact/maps4all/app/bulk_resource/views.py", line 588, in save_csv
req_opt_desc_const = RequiredOptionDescriptorConstructor.query.all()[0]
IndexError: list index out of range

- I think we need to deal with the case where we don't have an option descriptor, and therefore don't set a required option descriptor. Rn it's querying for it even when it isn't there i believe
 (in this case, I did reset -> 2 text descriptors)

app/templates/bulk_resource/review_desc_options.html
anniemeng: if this is an update and a descriptor already exists in the app it's shown below in the Existing Option Descriptor Values in CSV starting from line 65 (where i show what values are in the CSV and what are in the app) - so do we need to say that CSV adds to the options?

anniemeng:
i just merged master - should be fixed now

anniemeng:
also re the `Cathy's house` resource, it's because if you look at the apostrophe, it's a formatting thing where the second one is a bit funky, i think the csv has some extra styling on it

app/templates/bulk_resource/review_desc_options.html
riyer15: yeah  i think so! i dont think it's clear

riyer15:
lgtm!!

PR #146, opened 2017-02-03
Removing psycopg2 from requirements.

tvieira:
As far as I tested and run the application locally, it doesn't need to have the postgresql library installed. It uses sqlite.

Also, if this is kept in the requirements, the user has to install the development library for its operating system. I think the if it can use sqlite for development, this requirement can be removed safely.

### DIFF
--- a/requirements/common.txt
+++ b/requirements/common.txt
@@ -27,7 +27,6 @@ Mako==1.0.1
 MarkupSafe==0.23
 mccabe==0.4.0
 pep8==1.7.0
-psycopg2==2.6.1
 pyflakes==1.0.0
 pytz==2016.3
 ratelim==0.1.6


anniemeng:
We use postgresql when we deploy to Heroku or another server provider and we directly connect the repo to the Heroku app so we need psycopg2 for that!

tvieira:
Thanks for your feedback @anniemeng - to install psycopg2 python library I also have to install postgresql libraries from the distribution (brew for mac, or dnf for Fedora, and so on) - even though it doesn't need it for development but for your heroku deployment only. But I understood your point. Thanks for the great app.

abhisuri97:
Note that psycopg2 is not needed!

PR #158, opened 2018-02-04
Migrate to Python 3.6

katiejiang:
* Update README with .env variables and python3 virtual environment instructions
* Replace `geopy` with `googlemaps` since `geopy` doesn't work with Python 3.6
* Change `FILEPICKER_API_KEY` to `FILESTACK_API_KEY`
* Remove `requirements` folder and replace with `requirements.txt`


README.md
```diff
@@ -112,19 +127,19 @@ $ python manage.py add_fake_data
 ## Running the app
 
 ```
-$ source env/bin/activate
-$ foreman start -f Local
+$ source venv/bin/activate
+$ honcho start -f Local
 ```
 
 **Note**: if you are using Redis container with docker, you can ignore the error that `foreman` will display when you run the `foreman` command above. `foreman` will try to start the redis-server locally and it will not find the `redis-server` command. This is fine as long as you followed the [instructions above](#other-dependencies-for-running-locally) to run Redis in a container. The error should look like this
```
HanaPearlman: Can we remove this?

katiejiang: Yep!

PR #159, opened 2018-02-10
Remove latitude/longitude fields from resource form/table

katiejiang:
* Hide latitude/longitude form fields in `SingleResourceForm`
* Remove latitude/longitude columns in resource table


katiejiang:
Small bug with suggestions with latitude/longitude removal, I'll fix this!

PR #160, opened 2018-02-10
Allow admins to update Twilio credentials

katiejiang:


katiejiang:
Need to make sure it doesn't still import from os, testing.

katiejiang:
This works!

PR #161, opened 2018-02-11
update info about required columns in CSV upload

stephanieyshi:
just added this text "All resources must have a name under the 'Name' column and an address that can be converted by Google Maps into latitude/longitude under the 'Address' column."


stephanieyshi:
Is there any easy way we can maybe put up a sample CSV that people can download?

stephanieyshi:
Going to merge this for now

PR #165, opened 2018-02-17
New "moderator" role

stephanieyshi:
- created a new moderator role: moderator can add/delete resources, add/delete descriptors, CSV upload (update only) (ideally would have moderator only can edit resources that it created, will be a future task)
- edited the routes to restrict the routes that the moderator can access
- edited the nav bar 
- i think we need to do a DB migration or something before we can merge this in


stephanieyshi:
image

stephanieyshi:
@HanaPearlman should we merge this in with everything else?

HanaPearlman:
This code looks good, although I'm wondering about the functionality where a Moderator can only edit the resources that they uploaded to the site?

stephanieyshi:
merging in!

PR #167, opened 2018-02-17
Map center

katiejiang:
* If no resources, initialize map center to center of US
* Otherwise, map is centered at epicenter of resources
* Issues with current location (lag) so removed that

### DIFF
--- a/app/assets/scripts/map.js
+++ b/app/assets/scripts/map.js
@@ -262,46 +262,34 @@ function getCurrentLocation(callback) {
  * Initializes the map, the corresponding list of resources and search
  * functionality on the resources
  */
-function initMap(callback) {
+function initMap() {
   // Hide resource-info
   $("#resource-info").empty();
   $("#resource-info").hide();
 
-  getCurrentLocation(function(success, currentLocation) {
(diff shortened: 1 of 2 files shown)


katiejiang:
This is ready to merge @HanaPearlman

app/assets/scripts/map.js
```diff
@@ -480,17 +495,18 @@ function populateMarkers(resources) {
...
     oms.addMarker(markers[i]);
     bounds.extend(markers[i].getPosition());
   }
-
+  allResourceBounds = bounds;
   map.fitBounds(bounds);
   map.setCenter(bounds.getCenter());
-  allResourceBounds = bounds;
+  if (markers.length == 1) {
+    map.setZoom(15);
```
HanaPearlman: this is so small but: thoughts on zooming this out a little farther? the super close zoom was a bit disorienting for me.

katiejiang: On it!

katiejiang: Actually this zoom level is good, there's some other weird issue right now, I'll look into it

PR #169, opened 2018-02-17
Make latitude/longitude fields read-only

katiejiang:


katiejiang:
This is ready to merge @HanaPearlman

stephanieyshi:
I agree with Hana, we should do that

katiejiang:
Closing this because these changes are under #170

PR #171, opened 2018-02-19
make contact editor name and page name match

HanaPearlman:
page_name field is unique, and this was a copy paste that never got changed, so it was causing the integrity error


stephanieyshi:
LGTM

PR #173, opened 2018-03-08
Hyperlink Descriptors

katiejiang:


katiejiang:
Changes the database, so consider merging later!

app/main/views.py
```diff
@@ -125,7 +127,7 @@ def get_associations(resource_id):
             values = set(curr)
```
HanaPearlman: The previous line should be changed to curr['value'].append(val) since now "curr" is an object not just a list!

HanaPearlman:
One more thing-- when I add_fake_data and then look at advanced options, I can see one called "Website." So it seems like Hyperlinks are being included by accident in the advanced options, as if they were option descriptors.

katiejiang:
ok so I implemented a really quick fix to change the color of the hyperlink but it's probably horrible style pls help me @stephanieyshi :'(

stephanieyshi:
Can you merge in master? I'm seeing this 😢 
image

katiejiang:
@stephanieyshi I think the issue is that you have the old maps.js code cached on your browser. Try going to history and clearing your cache, and reload the page!

katiejiang:
screen shot 2018-03-31 at 4 12 33 pm

HanaPearlman:
This is good to merge! Not sure where we stand on merging changes which update the db but this is great! Thanks for making the changes.

PR #175, opened 2018-03-12
fix create pages page layout, fix flashes, update table if page is added

stephanieyshi:


HanaPearlman:
We should include some error checking to make sure the page _name_ don't already exist before someone creates them (we're checking the page url but not the name). 
We should also include an indication of which characters are allowed for the page url/what the page url is/how a user can access the page (eg, www.maps4allinstance.com/pages/pageurl)!

PR #177, opened 2018-03-19
added twilio authentication check

robertaninfeliz:


stephanieyshi:
So the "Text Me" hides on the front end, but when I add Twilio keys, it doesn't appear when I refresh. I can't tell if this is an issue with @katiejiang's form for adding Twilio Keys or the Jinja

app/main/views.py
```diff
@@ -33,7 +33,10 @@ def index():
     if req_opt_desc:
         for val in req_opt_desc.values:
             req_options[val] = False
-    return render_template('main/index.html', options=options_dict, req_options=req_options, req_desc=req_opt_desc)
+    twilio_auth = False
+    if os.environ.get('TWILIO_AUTH_TOKEN') is not None:
```
HanaPearlman: Hi! Things have been changed -- instead of just checking os.environment, you also have to check TwilioAuth in SiteAttributes! Sry!!

PR #179, opened 2018-03-19
Delete resources

robertaninfeliz:
screen shot 2018-03-19 at 7 04 30 pm

This what the alertbox looks like right now. Do we want more styling?


stephanieyshi:
Thoughts on using a modal? https://semantic-ui.com/modules/modal.html 
@katiejiang @HanaPearlman

robertaninfeliz:
Also, was thinking of using this:

https://sweetalert.js.org/docs/#configuration

katiejiang:
Ooh that's such a pretty alert! The only problem is that it doesn't exactly match our style, but I think the difference is small, so would be down to either use that or use a Semantic UI modal @stephanieyshi @robertaninfeliz

robertaninfeliz:
screen shot 2018-03-19 at 10 16 16 pm

katiejiang:
This is really cool! I would switch the order of Yes and Cancel, and change the text from "Yes" to "Yes, delete this resource" and make the color red, and make the color of the cancel button grey. Or make both grey!

robertaninfeliz:
screen shot 2018-03-19 at 11 27 42 pm

stephanieyshi:
Modal looks good! 

- I think it might be helpful to add a flash in the redirect after the resource is deleted to let them know that their resource has been successfully deleted.
- Could you also move the position of the delete button? Feel free to use whatever you think is best! Can also consider using a Semantic UI divider between the "Save Resource" button and the "Delete Resource" button.

PR #182, opened 2018-03-23
Download resources as CSV

katiejiang:
### Before
screen shot 2018-03-23 at 4 37 10 pm

### After
screen shot 2018-03-23 at 4 36 21 pm
screen shot 2018-03-23 at 4 36 39 pm
screen shot 2018-03-23 at 4 44 30 pm

## CSV Download
screen shot 2018-03-23 at 5 49 53 pm

### DIFF
--- a/app/main/views.py
+++ b/app/main/views.py
@@ -107,31 +107,16 @@ def search_resources():
     resources_as_dicts = Resource.get_resources_as_dicts(resources)
     return json.dumps(resources_as_dicts)
 
+
 @main.route('/get-associations/<int:resource_id>')
 def get_associations(resource_id):
     resource = Resource.query.get(resource_id)
-    associations = {}
-    if resource is None:
-        return json.dumps(associations)
-    for td in resource.text_descriptors:
-        associations[td.descriptor.name] = td.text
-    for od in resource.option_descriptors:
-        val = od.descriptor.values[od.option]
-        values = set()
-        # multiple option association values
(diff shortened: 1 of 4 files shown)


app/main/views.py
```diff
@@ -107,31 +107,16 @@ def search_resources():
...
-            values.add(val)
-        associations[od.descriptor.name] = list(values)
-    return json.dumps(associations)
+    return json.dumps(Resource.get_associations(resource))
 
 
-@main.route('/pages/<pageName>')
-def render_page(pageName):
-    editable_html_obj = EditableHTML.get_editable_html(pageName)
+@main.route('/pages/<page_name>')
```
stephanieyshi: Can we put these edits in a separate PR? Also this is causing a build error:
image
(Works if I recreate the DB)

stephanieyshi: ok NVM about putting it in a separate PR, but is there a way to fix this BuildError without having to recreate the db

katiejiang:
@stephanieyshi I actually tried that, but then I realized that it's a good idea to have the button there so users know its an option.

app/main/views.py
katiejiang: Tbh this was just a style fix that bothered me, I just changed it back bc it's not super important!

HanaPearlman:
LGTM, this is good to merge. Got an error when I first ran the branch ("str has no attribute append") but when I recreated the db, added fake data, and ran again it was perfect.

PR #184, opened 2018-03-23
Unique descriptor name validator

katiejiang:


stephanieyshi:
Can we add a note like "Descriptor names must be unique and are case sensitive." on the "Add New Descriptor" page? Otherwise LGTM

stephanieyshi:
LGTM! Merging

PR #188, opened 2018-04-10
Fix download resource bug

katiejiang:
Also changes `add_fake_data` to add option descriptors that make more sense!


katiejiang:
Fixes the bug where having multiple option descriptor screws up the number of columns in a resource row!

PR #191, opened 2018-04-11
Fix NoneType error when editing text descriptor

katiejiang:
NoneType error when trying to edit a text descriptor from "All Descriptors" view


stephanieyshi:
does this address the error message i forwarded to you?

katiejiang:
I don’t think it’s the same bug, that one has to do with the missing dtype!

PR #192, opened 2018-04-11
Download uses semicolons to separate option descriptor values

katiejiang:


katiejiang:
screen shot 2018-04-11 at 2 12 06 pm

PR #194, opened 2018-04-11
add dtype to descriptors with db migration

HanaPearlman:
sets up a potential flow that we might want to use for future migrations! 
1) migrations folder contains a list of files, each specifying changes which should be made for a given migration
2) locally, developer should make db changes, run a "migrate" command, and edit the file which is produced as a result so it does exactly what we want it to do
3) test locally, and probably w a test heroku instance also, and then merge w master
4) in the Procfile the specified command "release" will be run for a heroku instance linked with github auto deploys; this command runs db upgrade, which attempts to perform the migration. 
as far as i've seen, if a migration has been  performed before, it will not re-run? could test more


katiejiang:
!!! This is awesome!

stephanieyshi:
woopiee!! let's go over this in the meeting today and i (and maybe someone else??) will review this after

HanaPearlman:
update i really don't think an upgrade is run twice, so it's fine to specify "db upgrade" in the release command i think!!

stephanieyshi:
we should also probably update the readme with information about migrations at some point

migrations/versions/4856a6d821b_add_dtype_to_descriptors.py
```diff
@@ -32,6 +32,7 @@ def upgrade():
         	descriptor.dtype = "text"
 
     session.commit()
+    print("done; silly change")
```
katiejiang: remove this before merging 😛

PR #195, opened 2018-04-18
Welcome modal

katiejiang:


stephanieyshi:
Should we call it modal? Or something else like popup 🤔

PR #198, opened 2018-04-27
Hana katie migration

HanaPearlman:
omg


HanaPearlman:
sets up a potential flow that we might want to use for future migrations!

migrations folder contains a list of files, each specifying changes which should be made for a given migration
locally, developer should make db changes, run a "migrate" command, and edit the file which is produced as a result so it does exactly what we want it to do
test locally, and probably w a test heroku instance also, and then merge w master
in the Procfile the specified command "release" will be run for a heroku instance linked with github auto deploys; this command runs db upgrade, which attempts to perform the migration.
as far as i've seen, if a migration has been performed before, it will not re-run? could test more

PR #202, opened 2018-05-19
Update README.md

dscotese:
I examined the project here on GitHub and didn't realize what I should do.  The Readme seems to have been written under the assumption that the only way to get to it is by logging into maps4all.org.  Since I didn't do that, I figure others also might not do it, so the text I added will help them proceed.


dscotese:
This request should be rejected and replaced with the other one I made.

